PR #1, opened 2022-02-03
Add application scaffolding 

jemoreira:


controller.go
```diff
@@ -0,0 +1,40 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
```
ser-io, line 7: nit: fix indentation.

jemoreira: Done

controller.go
```diff
@@ -15,26 +15,120 @@
...
+	"strconv"
+
+	"github.com/gorilla/mux"
 )
 
 // The controller implements the web API of the cloud orchestrator. It parses
 // and validates requests from the client and passes the information to the
 // relevant modules
 type Controller struct {
+	ss SignalingServer
```
ser-io, line 31: nit: should we have more descriptive field names? Go advocates for one letter names when scope is smaller, but field names could be used anywhere.

controller.go
```diff
@@ -126,6 +117,19 @@ func indexHandler(w http.ResponseWriter, r *http.Request) {
 	fmt.Fprintln(w, "Home page")
 }
 
+// Get an int form value from the request
```
ser-io, line 120: nit: let's make clear that `default` is only used if the field name does not exist, if it exists and it's an invalid value an error will be returned.

controller.go
jemoreira: absolutely!

controller.go
jemoreira: Done

ser-io (approved):

PR #2, opened 2022-02-14
Implement the main endpoints of the Signaling Server

jemoreira:


types.go
```diff
@@ -27,8 +31,28 @@ type ForwardMsg struct {
...
+	// response from the Host Orchestrator and the status code if it was
+	// able to communicate with it, otherwise it returns an error.
+	NewConnection(msg NewConnMsg) (SServerResponse, error)
+	Forward(id string, msg ForwardMsg) (SServerResponse, error)
+	Messages(id string, start int, count int) (SServerResponse, error)
+}
+
+type DeviceDesc struct {
+	// The (internal) address of the host where the cuttlefish device is running
+	Addr string
```
ser-io, line 50: let's add an example to have an idea about the format these address values will have.

types.go
```diff
@@ -27,8 +31,28 @@ type ForwardMsg struct {
...
+	// The id under which the cuttlefish devcie is registered with the host
```
ser-io, line 51: nit: device

signalingserver_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"testing"
+)
+
+func TestIdEncodingAndDecoding(t *testing.T) {
+	connId, devId := "connId", "devId"
+	AuxTestIdEncodingAndDecoding(connId, devId, t)
+}
+
```
ser-io, line 25: can we use the table-driven ( https://github.com/golang/go/wiki/TableDrivenTests) pattern here? it is encouraged in Go!

signalingserver.go
```diff
@@ -14,23 +14,161 @@
...
+	ConnId string
+	DevId  string
 }
 
-func (s *ForwardingSignalingServer) Messages(id string, start int, count int) ([]interface{}, error) {
-	return make([]interface{}, 0), nil
+func decodeConnId(connId string) (IdDecodeResult, error) {
+	idx := strings.LastIndex(connId, SEPARATOR)
+	if idx < 0 {
+		return IdDecodeResult{}, errors.New("Malformed connection id")
```
ser-io, line 166: let's mention that a separator is missing and add the connId value.

signalingserver.go
```diff
@@ -14,23 +14,161 @@
...
+		err = dec.Decode(resObj)
+	}
+	if err != nil {
+		log.Println("Failed to parse response json: ", err)
+		return -1, err
+	}
+	return res.StatusCode, nil
+}
+
+const SEPARATOR string = ":"
```
ser-io, line 149: suggest to add a CONN_ID prefix as SEPARATOR only is too generic, even for this context only.

signalingserver.go
```diff
@@ -14,23 +14,161 @@
...
+func (s *ForwardingSignalingServer) NewConnection(msg NewConnMsg) (SServerResponse, error) {
+	dev, err := s.instanceManager.DeviceFromId(msg.DeviceId)
+	if err != nil {
+		return SServerResponse{StatusCode: http.StatusInternalServerError}, err
+	}
+	var resErr ErrorMsg
+	var reply NewConnReply
+	status, err := POSTRequest(hostURL(dev.Addr, "/polled_connections", ""), NewConnMsg{dev.LocalId}, &reply, &resErr)
+	if err != nil {
+		return SServerResponse{StatusCode: status}, err
```
Databean, line 48: Is it possible to use [error wrapping](https://go.dev/doc/go1.13#error_wrapping) to attach more information to the errors returned from these functions? It looks like of the error handling here logs the extra information and then returns the original error. Augmenting the original error would allow constructing a single error object with all the relevant context. See also the [go 1.13 errors blog post](https://go.dev/blog/go1.13-errors).

signalingserver.go
```diff
@@ -14,23 +14,161 @@
...
+		query = fmt.Sprintf("%s&count=%d", query, count)
+	}
+	var resErr ErrorMsg
+	var reply interface{}
+	status, err := GETRequest(hostURL(dev.Addr, "/polled_connections/"+connId+"/messages", query), &reply, &resErr)
+	return SServerResponse{reply, status}, err
+}
+
+func hostURL(addr string, path string, query string) string {
+	url := "http://" + addr + ":1080" + path
```
Databean, line 98: The android-cuttlefish project promises this port number is [configurable](https://github.com/google/android-cuttlefish/blob/0861e8d9e6f2dee7f6070fb20dde98d7eb47dce4/debian/cuttlefish-common.default#L42) but here it's hardcoded, is this temporary?

types.go
jemoreira: Modified the comment to be more explicit about it returning a network address instead of a URL and clarified that either domain names and IP addresses (v4 and v6) are OK.

types.go
jemoreira: Done

signalingserver_test.go
jemoreira: Done

signalingserver.go
jemoreira: Done

signalingserver.go
jemoreira: Done

dummy.go
```diff
@@ -0,0 +1,24 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+// Implements the InstanceManager interface providing access to the first
+// device in the local host orchestrator.
+// This implementation is useful for both development and testing
+type DummyIM struct{}
```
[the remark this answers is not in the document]
jemoreira: Done

controller.go
```diff
@@ -70,8 +71,13 @@ func (c *Controller) CreateConnection(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	log.Println("id: ", msg.DeviceId)
-	reply := c.sigServer.NewConnection(msg)
-	replyJSON(w, reply)
+	reply, err := c.sigServer.NewConnection(msg)
+	if err != nil {
+		log.Println("Failed to send message to device: ", err)
```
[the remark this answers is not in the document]
jemoreira: Done
A status code still needs to be returned from each signaling server endpoint because the host orchestrator's reply includes one. IOW If the host orchestrator returns a non-2xx status code and a JSON object, both need to be returned to the client.

signalingserver.go
jemoreira: Yes, this is temporary. I need to figure out what's the best way to configure the app. That configuration will include the port used by the host orchestrator, the protocol and possible some auth metadata.

signalingserver.go
jemoreira: Yes, this is temporary. I need to figure out what's the best way to configure the app. That configuration will include the port used by the host orchestrator, the protocol and possible some auth metadata.

ser-io (approved):

signalingserver.go
jemoreira: Done

Databean (approved):

PR #5, opened 2022-02-22
Add the remaining signaling server endpoints

jemoreira:
As of this change a connection to a device can be established through the cloud orchestrator's signaling server:
* Start a device on the same host where the orchestrator is running, with name cvd-1 (the default)
* On a browser, navigate to http://<Address>/device/cvd-1/files/client.html?deviceId=cvd-1


intercept/js/server_connector.js
```diff
@@ -0,0 +1,182 @@
...
+    this.#messagesReceived += arr.length;
+    return arr;
+  }
+
+  #startPolling() {
+    if (this.#pollerSchedule !== undefined) {
+      return;
+    }
+
+    let currentPollDelay = 1000;
```
ser-io, line 166: should we declare a constant for the initial poll delay?

ser-io (approved):

Databean (approved):

intercept/js/server_connector.js
jemoreira: we could, but since it won't be used anywhere else and 1000 is quite commonly used as 1 second in milliseconds, especially in javascript, I thought it wouldn't make the code much clearer than just using the number directly.

PR #6, opened 2022-03-15
AccountManager implementation

jemoreira:
Two implementations are available:
- One for local development, that assumes the user is always logged in and gets the username from the environment
- One for running in GAE, which gets the username from the HTTP request header.
The application selects between the two implementations at runtime by checking for the presence of GAE specific environment variablees.


gaeusersaccountmanager.go
```diff
@@ -0,0 +1,59 @@
...
+func (g *GAEUsersAccountManager) Authorize(fn AuthHTTPHandler) HTTPHandler {
+	return func(w http.ResponseWriter, r *http.Request) error {
+		// These headers are guaranteed to be present and come from AppEngine.
+		email := r.Header.Get(emailHeaderKey)
+		if email == "" {
+			// Normally a redirect to the login page is returned here, but the App Engine
+			// api takes care of that so if a request gets this far without the headers
+			// it can only be due to server misconfiguration.
+			// A general error returned here will trigger a 500 response
+			return errors.New("No authentication headers present")
```
ser-io, line 41: nit: should we add the specific header that's expected?

gaeusersaccountmanager.go
```diff
@@ -0,0 +1,59 @@
...
+		}
+		username := r.Header.Get(usernameHeaderKey)
```
ser-io, line 43: can the same person be authenticated with two different usernames? a first time passing a nickname and a second time not passing the nickname therefore using the email?

types.go
```diff
@@ -47,14 +47,18 @@ type IceServer struct {
...
 	// All endpoints in the SignalingServer return the (possibly modified)
 	// response from the Host Orchestrator and the status code if it was
 	// able to communicate with it, otherwise it returns an error.
-	NewConnection(msg NewConnMsg) (*SServerResponse, error)
-	Forward(id string, msg ForwardMsg) (*SServerResponse, error)
-	Messages(id string, start int, count int) (*SServerResponse, error)
-	ServeDeviceFiles(devId string, path string, w http.ResponseWriter, r *http.Request) error
+	NewConnection(msg NewConnMsg, user UserInfo) (*SServerResponse, error)
+	Forward(id string, msg ForwardMsg, user UserInfo) (*SServerResponse, error)
+	Messages(id string, start int, count int, user UserInfo) (*SServerResponse, error)
```
ser-io, line 60: should we encapsulate this parameters into a new MessagesRequest struct to avoid the long list of parameters, the same for Forward and ServeDeviceFiles?

osaccountmanager.go
```diff
@@ -0,0 +1,41 @@
...
+package main
+
+import (
+	"net/http"
+	"os"
+)
+
+// Implements the AccountManager interface taking the username from the
+// environment and authorizing all requests
+type OsAccountManager struct{}
```
ser-io, line 24: is this account manager meant for development purposes only? if so, let's add it to the documentation.

types.go
```diff
@@ -67,5 +71,18 @@ type DeviceDesc struct {
...
+type HTTPHandler func(http.ResponseWriter, *http.Request) error
+
+type AccountManager interface {
+	// Returns the received http handler wrapped in another that extracts user
+	// information from the request and passes it to to the original handler as
+	// the last parameter.
+	// The wrapper will only pass the request to the inner handler if a user is
+	// authenticated, otherwise it may choose to return an error or respond with
+	// an HTTP redirect to the login page.
+	Authorize(fn AuthHTTPHandler) HTTPHandler
```
ser-io, line 87: Authenticate instead? Authorize is more related to permissions over resources.

gaeusersaccountmanager.go
```diff
@@ -0,0 +1,59 @@
...
+package main
+
+import (
+	"errors"
+	"net/http"
+	"strings"
+)
+
+const emailHeaderKey = "X-Appengine-User-Email"
+const usernameHeaderKey = "X-Appengine-User-Nickname"
```
ser-io, line 24: nit: use a single const declaration for related values.

gaeusersaccountmanager.go
```diff
@@ -0,0 +1,59 @@
...
+
+type GAEUsersAccountManager struct{}
+
+func NewGAEUsersAccountManager() *GAEUsersAccountManager {
+	return &GAEUsersAccountManager{}
+}
+
+func (g *GAEUsersAccountManager) Authorize(fn AuthHTTPHandler) HTTPHandler {
+	return func(w http.ResponseWriter, r *http.Request) error {
+		// These headers are guaranteed to be present and come from AppEngine.
```
Databean, line 34: Would it be possible to use the official [appengine package](https://pkg.go.dev/google.golang.org/appengine) here? It includes a support library for [user information](https://pkg.go.dev/google.golang.org/appengine@v1.6.7/user).

types.go
jemoreira: I think Authorize makes more sense here:
We are not trying to identify in a secure way who the user really is (authentication), GAE takes care of that for us, but instead we're deciding if a user should be allowed to execute this request (authorization). It just so happens that all (authenticated) users are allowed to do all operations for now. In the future we could extend this function to take the handler and a role that is allowed to make the request so it would check that the user has that role.

osaccountmanager.go
jemoreira: We'll use it for that purpose only, but others are free to use it for something else if it makes sense.

gaeusersaccountmanager.go
jemoreira: Yes, but I chose not to because of the following in no particular order:
- I don't want to make it too easy to add even more dependencies on app engine
- it's easier (in my opinion) to get the information from the headers than to create a dependency with a library and then use it
- we're not exposed to bugs or security issues from the library (unlikely to affect us for such a simple use, but not impossible)
- if app engine were to do a breaking change (change the header or stop using it altogether) we'd still be affected by it because we'd need to update the version of the library

gaeusersaccountmanager.go
jemoreira: That string may end up in two places:
- logs: makes sense to have the header there, but it's not hard to find if it's not: Anyone with access to the logs should have access to the source as well.
- returned to the user: I'd rather not hint at the user what header is missing. If we reach this error it either means our server is misconfigured and the missing header won't help the user or the user is doing something bad and we shouldn't try to help them
Given the above I think it's best not to include the missing header.

types.go
jemoreira: Did for SeveDeviceFiles.

Messages and Forward only have 4 and 3 parameters each and the lines are not that long

gaeusersaccountmanager.go
jemoreira: I don't think so, since these are google accounts and those don't let you specify a nickname. But no need to risk it, I switched to always get it from the email.

gaeusersaccountmanager.go
Databean: What I'm afraid of here is rewriting the app engine package through a series of "one more thing" changes. If the issue is that you want to keep the option open for non-appengine implementations of AccountManager, would it be too much additional work to figure out what the equivalent would be for e.g. AWS to validate that this is extensible?

types.go
Databean: All it's doing now is still only authentication, and the API doesn't look like it's really set up to make authorization convenient. As soon as the authorization gets more complex than "user X can access all endpoints and user Y cannot", this would either have to reimplement the request routing table to discover which per-request logic it has to run, or the request handler has to do its own authorization, again leaving this as purely authentication (whether relying on app engine information or otherwise).

.gcloudignore
```diff
@@ -0,0 +1,25 @@
...
+# Binaries for programs and plugins
+*.exe
+*.exe~
+*.dll
+*.so
+*.dylib
+# Test binary, build with `go test -c`
+*.test
+# Output of the go coverage tool, specifically when used with LiteIDE
+*.out
```
Databean, line 25: nit: newline

Databean (approved):

types.go
jemoreira: The one thing this api is missing to be obviously authorization is an extra parameter of type UserRole (one of ADMIN, VIEWER, EDITOR, for example). The role parameter would be passed along with the http handler and the implementation of this method would check that the current user is the given role. I didn't add the extra parameter because we don't have a clear idea of what the roles will be yet.

Even without the extra parameter I still think this is authorization. My understanding of the subject can be summarized as:
"authentication is the process of verifying who someone is , whereas authorization is the process of verifying what things a user has access to"
It's impossible to do authorization without doing authentication first, so it's very easy to mix the two, however this method is "deciding whether a request should be allowed to proceed based on who the user is" which, in my opinion, matches the concept of authorization better than it does with authentication.

That being said, I could be wrong. Given that both of you seem to agree I'll rename it, we can continue the debate on another time and rename back if we all agree to do so.

gaeusersaccountmanager.go
jemoreira: This code is not trying to be extensible, it's doing a very app engine specific thing.

"rewriting the app engine package through a series of "one more thing" changes"
That sounds bad, but is it really?
- The header approach is the only option for some runtimes (example nodejs), so it's unlikely to change.
- The code for getting the username from the app is not going to be simpler than this (call a function versus get a value from a dictionary) and will require additional configuration and dependency handling.

If another dependency on app engine needs to be added later we can decide whether it makes sense to add the library then or go this route again.

ser-io (approved):

PR #7, opened 2022-03-26
Creates host instance in GCP.

ser-io:


gcpim.go
```diff
@@ -0,0 +1,121 @@
...
+	"errors"
+	"github.com/google/uuid"
+	computepb "google.golang.org/genproto/googleapis/cloud/compute/v1"
+	"google.golang.org/protobuf/proto"
+)
+
+const (
+	// TODO(b/220891296): Make this configurable
+	projectId   = "google.com:cloud-android-jemoreira"
+	sourceImage = "projects/cloud-android-releases/global/images/cuttlefish-google-vsoc-0-9-21"
```
jemoreira, line 16: These names are internal and should not be available in an open source repo. We're probably going to have to solve design the configuration feature before submitting this PR.

gcpim.go
```diff
@@ -0,0 +1,121 @@
...
+	networkName = "projects/cloud-android-jemoreira/global/networks/default"
+)
+
+const (
+	namePrefix  = "cf-"
+	labelPrefix = "cf-"
+)
+
+var newUUIDString = func() string {
```
jemoreira, line 25: This should be a function instead of a variable

gcpim.go
```diff
@@ -0,0 +1,121 @@
...
+	return uuid.New().String()
+}
+
+// GCP implementation of the instance manager.
+type GCPIM struct {
```
jemoreira, line 30: nit: GCPInstanceManager

gcpim.go
```diff
@@ -0,0 +1,121 @@
...
+		Zone:    zone,
+		InstanceResource: &computepb.Instance{
+			Name:           proto.String(namePrefix + newUUIDString()),
+			MachineType:    proto.String(req.HostInfo.GCP.MachineType),
+			MinCpuPlatform: proto.String(req.HostInfo.GCP.MinCPUPlatform),
+			Disks: []*computepb.AttachedDisk{
+				{
+					InitializeParams: &computepb.AttachedDiskInitializeParams{
+						DiskSizeGb:  proto.Int64(int64(req.HostInfo.GCP.DiskSizeGB)),
+						SourceImage: proto.String("projects/cloud-android-releases/global/images/cuttlefish-google-vsoc-0-9-21"),
```
jemoreira, line 61: use the constant

gcpim.go
```diff
@@ -0,0 +1,121 @@
...
+					AccessConfigs: []*computepb.AccessConfig{
+						{
+							Name: proto.String("External NAT"),
+							Type: proto.String(computepb.AccessConfig_ONE_TO_ONE_NAT.String()),
+						},
+					},
+				},
+			},
+			Labels: map[string]string{
+				labelPrefix + "creator":  user.Username(),
```
jemoreira, line 78: add the label used by acloud too (created_by), it's ok to have both creator and created_by, though we'd probably be better sticking with acloud's name only.

gcpim.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (m *GCPIM) DeviceFromId(name string, _ UserInfo) (DeviceDesc, error) {
+	return DeviceDesc{"127.0.0.1", "cvd-1"}, nil
+}
+
+func (m *GCPIM) InsertHost(zone string, req *imtypes.InsertHostRequest, user UserInfo) (*imtypes.Operation, error) {
+	if err := validateRequest(req); err != nil {
+		return nil, err
+	}
+	ctx := context.Background()
```
jemoreira, line 49: From https://pkg.go.dev/context#TODO it seems context.TODO() is a better match for this.

main.go
```diff
@@ -15,16 +15,25 @@
...
 
 func main() {
-	im := &PlaceholderIM{}
+	ctx := context.Background()
+	client, err := compute.NewInstancesRESTClient(ctx)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer client.Close()
+	im := NewGCPIM(client)
```
jemoreira, line 36: encapsulate all this in a New....() function.

jemoreira (changes requested):

api/instancemanager/v1/types.go
```diff
@@ -0,0 +1,106 @@
+package v1
```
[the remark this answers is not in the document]
ser-io: handled on https://github.com/google/cloud-android-orchestration/pull/8

api/instancemanager/v1/types.go
```diff
@@ -0,0 +1,106 @@
...
+
+type InsertHostRequest struct {
```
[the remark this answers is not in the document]
ser-io: changed it.

api/instancemanager/v1/types.go
```diff
@@ -0,0 +1,106 @@
...
+	HostInfo *HostInfo `json:"host_info"`
+}
+
+type CVDInfo struct {
+	// [REQUIRED] A number that uniquely identifies the a set of builds of different targets.
+	BuildID string `json:"build_id"`
+	// [REQUIRED] A string to determine the specific product and flavor from the set of builds, e.g. aosp_cf_x86_64_phone-userdebug
+	Target string
+	// The number of CVDs to create. Use this field if creating more than one instance.
+	InstancesNumber int `json:"instances_number"`
```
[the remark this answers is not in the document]
ser-io: done

api/instancemanager/v1/types.go
```diff
@@ -0,0 +1,106 @@
...
+
+type HostInfo struct {
+	// Required if using GCP.
+	GCP *GCPInstance `json:"gcp"`
+}
+
+type GCPInstance struct {
+	// [REQUIRED]
+	DiskSizeGB int `json:"disk_size_gb"`
+	// [REQUIRED] More info about this field here for https://cloud.google.com/compute/docs/reference/rest/v1/instances/insert#request-body
```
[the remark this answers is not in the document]
ser-io: done

api/instancemanager/v1/types.go
```diff
@@ -0,0 +1,106 @@
...
+	ResultObject interface{}
+}
+
+type Error struct {
+	Code    string
+	Message string
+}
+
+//
+// Helper methods to easy how to test nil or default zero values in nested structs.
```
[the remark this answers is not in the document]
ser-io: agree, removed the helper methods.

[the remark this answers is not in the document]
ser-io: removed the helper methdods.

controller.go
```diff
@@ -60,6 +62,9 @@ func (c *Controller) SetupRoutes() {
 	router.Handle("/connections", HTTPHandler(c.accountManager.Authenticate(c.CreateConnection))).Methods("POST")
 	router.Handle("/devices/{deviceId}/files{path:/.+}", HTTPHandler(c.accountManager.Authenticate(c.GetDeviceFiles))).Methods("GET")
 
+	// Instance Manager Routes
+	router.Handle("/v1/zones/{zone}/hosts", HTTPHandler(c.accountManager.Authenticate(c.InsertHost))).Methods("POST")
```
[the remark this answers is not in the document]
ser-io: done in https://github.com/google/cloud-android-orchestration/pull/8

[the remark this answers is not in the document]
ser-io: not in this cl, but in the coming one.

gcpim.go
ser-io: done

gcpim.go
ser-io: done, kept both looking forward to remove the the acloud one once deprecated and have all the labels with a specific prefix, `cvd` in this case.

gcpim.go
ser-io: why TODO will be a better match here? documentation related to GCP client libraries always use context.Background() in their how-to documentation when making requests, an empty context could cause the requests to fail.

main.go
ser-io: done

gcpim.go
ser-io: done

api/instancemanager/v1/types.go
```diff
@@ -0,0 +1,106 @@
...
+	// [REQUIRED]
+	DiskSizeGB int `json:"disk_size_gb"`
+	// [REQUIRED] More info about this field here for https://cloud.google.com/compute/docs/reference/rest/v1/instances/insert#request-body
+	MachineType string `json:"machine_type"`
+	// More info about this field here for https://cloud.google.com/compute/docs/reference/rest/v1/instances/insert#request-body
+	MinCPUPlatform string `json:"min_cpu_platform"`
+}
+
+type Operation struct {
+	Name string
```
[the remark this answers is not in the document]
ser-io: done

api/instancemanager/v1/types.go
```diff
@@ -0,0 +1,106 @@
...
+	// Service-specific metadata associated with the operation.  It typically
+	// contains progress information and common metadata such as create time.
+	Metadata interface{} `json:",omitempty"`
+	// If the value is `false`, it means the operation is still in progress.
+	// If `true`, the operation is completed, and either `error` or `response` is
+	// available.
+	Done bool
```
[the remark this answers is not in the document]
ser-io: it avoids enforcing to populate the underlying result object. some api might chose to populate the underlying result object whereas others not.

api/instancemanager/v1/types.go
```diff
@@ -0,0 +1,106 @@
...
+
+type InsertHostRequest struct {
+	// [REQUIRED]
+	CVDInfo *CVDInfo `json:"cvd_info"`
+	// [REQUIRED]
+	HostInfo *HostInfo `json:"host_info"`
+}
+
+type CVDInfo struct {
+	// [REQUIRED] A number that uniquely identifies the a set of builds of different targets.
```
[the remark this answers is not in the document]
ser-io: updated the documentation for this field to only "The Android build identifier."

gcpim.go
ser-io: Created a minimal config object that is passed to the Instance Manager with at least this two fields removing the values from the code as suggested.

api/instancemanager/v1/types.go
```diff
@@ -0,0 +1,106 @@
...
+	BuildID string `json:"build_id"`
+	// [REQUIRED] A string to determine the specific product and flavor from the set of builds, e.g. aosp_cf_x86_64_phone-userdebug
```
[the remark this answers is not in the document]
ser-io: kept these two required, however made the CreateCVDRequest field optional in the CreateHostRequest object.

api/instancemanager/v1/types.go
```diff
@@ -0,0 +1,106 @@
+package v1
+
+type InsertHostRequest struct {
+	// [REQUIRED]
+	CVDInfo *CVDInfo `json:"cvd_info"`
+	// [REQUIRED]
+	HostInfo *HostInfo `json:"host_info"`
+}
+
+type CVDInfo struct {
```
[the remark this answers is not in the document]
ser-io: done

controller.go
```diff
@@ -140,6 +145,25 @@ func (c *Controller) Forward(w http.ResponseWriter, r *http.Request, user UserIn
...
+	zone := mux.Vars(r)["zone"]
+	var msg imtypes.InsertHostRequest
+	err := json.NewDecoder(r.Body).Decode(&msg)
+	if err != nil {
+		return NewBadRequestError("Malformed JSON in request", err)
+	}
+	op, err := c.instanceManager.InsertHost(zone, &msg, user)
+	if err != nil {
+		if errors.Is(err, ErrBadInsertHostRequest) {
+			return NewBadRequestError("", err)
```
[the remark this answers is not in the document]
ser-io: used the new error api after merging https://github.com/google/cloud-android-orchestration/pull/9

controller.go
```diff
@@ -140,6 +145,25 @@ func (c *Controller) Forward(w http.ResponseWriter, r *http.Request, user UserIn
...
+		}
+		log.Printf("InsertHost error %v", err)
+		return NewInternalError("", nil)
```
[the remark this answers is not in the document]
ser-io: done, used new errors api after merging https://github.com/google/cloud-android-orchestration/pull/9

api/instancemanager/v1/types.go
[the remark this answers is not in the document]
jemoreira: One way forces to populate a result field, but it can populate it with just {} since everything inside it is omitempty.
The other way forces to "populate" a done field with either 'true' or 'false'.
Both approaches have the same number of "forces to populate". The only real difference I see is that the first one forces the client to check the result field is populated after it already checked done is 'true'.

gcpim.go
jemoreira: Both Background and TODO say: " returns a non-nil, empty Context.".

TODO says: "Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter)."

Background says: "It is typically used by the main function, initialization, and tests, and as the top-level Context for incoming requests."

It seems both do the same thing, but the one from Background is meant to be the first context ever created at the entry point (incoming requests, main function, etc) while TODO is meant to be used elsewhere when there is no surrounding context available, probably to make it easier to find once a surrounding context is available.

gcpim.go
jemoreira: we can do much better than having a non constant global variable. One option is to receive a uuidFactory of the same type as your variable as a parameter in the constructor, or even better to have it be an instance variable created with a default value in the constructor but replaceable through a setter function.

api/instancemanager/v1/types.go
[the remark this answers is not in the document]
ser-io: discussed offline and decided to keep the `done` field due clarity of use

gcpim.go
ser-io: this variable is only accessible within its package, which is how white-box testing is done in Golang, then the argument is about white-box testing vs black-box testing which is what you suggest. Now, context is key here to decide which way to go. Should we make passing a uuidFactory part of the public API and add the extra work that comes with it or can we keep it internally and do white-box testing? In this case specifically, I think passing and handling a uuidFactory is dirtying the public API as UUID generation could be kept internally. If having another use case different that testing to move it to the public API I won't argue against it.

I'm aware, this code uses the main package right now, I tried to have it on its own package but right now is impossible without moving other code likes errors.go and types.go into their own packages different than main, otherwise we have cyclic dependencies. I'd like the change of splitting main into different packages to be done in a different PR to keep this one focused on the creation process only.

gcpim.go
ser-io: good to know, thanks! done.

gcpim.go
jemoreira: You don't have to make it part of the public API, you can create a setter in the class that is not exported from the package the same way you're doing with the global variable.

With regards to white box vs black box testing: I prefer to add something to the public API to enable testing than to have my object access a mutable global variable. Mutable global variable are a known source of hard to debug bugs while an extra element in the API will only confuse the users a little and force them to read documentation.

ser-io: due the current lack of constraints on modifying the package level variable, we decided to move the uuid factory to the public api.

gcpinstancemanager.go
```diff
@@ -0,0 +1,137 @@
...
+// GCP implementation of the instance manager.
+type GCPInstanceManager struct {
+	config      *Config
+	uuidFactory func() string
+	client      *compute.InstancesClient
+}
+
+func NewGCPInstanceManager(
+	config *Config,
+	uuidFactory func() string,
```
jemoreira, line 43: This forces service users to provide a UUID factor. I think it's better to just create it with a default value here and provide a setter.

jemoreira (changes requested):

gcpinstancemanager.go
ser-io: done

jemoreira (approved):

PR #8, opened 2022-03-28
Creates package for the Signaling Server API public elements.

ser-io:
- Version, with prefix `v1`, the current services endpoints.


controller.go
```diff
@@ -22,9 +22,57 @@ import (
...
 	"strconv"
 
+	caoapi "cloud-android-orchestration/api/v1"
+
 	"github.com/gorilla/mux"
 )
 
+type ErrorMsg struct {
+	Error string `json:"error"`
+}
```
jemoreira, line 32: This should be in api/v1/signalingserver.go or in api/v1/common.go if you want to reuse it for something else. Feel free to rename it if it collides with something.

controller.go
```diff
@@ -22,9 +22,57 @@ import (
...
+
+type InfraConfig struct {
+	IceServers []IceServer `json:"ice_servers"`
+}
+type IceServer struct {
+	URLs []string `json:"urls"`
+}
```
jemoreira, line 39: These two are part of the API spec, so they should be under api/v1 too

controller.go
```diff
@@ -22,9 +22,57 @@ import (
...
+	// running. The address can either be an IPv4, IPv6 or a domain name.
+	Addr string
+	// The id under which the cuttlefish device is registered with the host
+	// orchestrator (can be different from the id used in the cloud orchestrator)
+	LocalId string
+}
+
+type InstanceManager interface {
+	DeviceFromId(name string, user UserInfo) (DeviceDesc, error)
+}
```
jemoreira, line 74: These interfaces don't belong in controller.go, they should be in a different file also under api/v1 (see the other comment about moving controller.go).

controller.go
```diff
@@ -22,9 +22,57 @@ import (
 	"net/http"
 	"strconv"
 
+	caoapi "cloud-android-orchestration/api/v1"
```
jemoreira, line 25: nit: rename as v1api or api1

controller.go
```diff
@@ -55,10 +103,10 @@ func (c *Controller) SetupRoutes() {
 	router := mux.NewRouter()
 
 	// Signaling Server Routes
-	router.Handle("/connections/{connId}/messages", HTTPHandler(c.accountManager.Authenticate(c.Messages))).Methods("GET")
-	router.Handle("/connections/{connId}/:forward", HTTPHandler(c.accountManager.Authenticate(c.Forward))).Methods("POST")
-	router.Handle("/connections", HTTPHandler(c.accountManager.Authenticate(c.CreateConnection))).Methods("POST")
-	router.Handle("/devices/{deviceId}/files{path:/.+}", HTTPHandler(c.accountManager.Authenticate(c.GetDeviceFiles))).Methods("GET")
+	router.Handle("/v1/connections/{connId}/messages", HTTPHandler(c.accountManager.Authenticate(c.Messages))).Methods("GET")
```
jemoreira, line 106: I think the controller.go file should be in api/v1 too. After all it's setting it up.

jemoreira (changes requested):

controller.go
ser-io: moved to signalingserver.go

controller.go
ser-io: moved to back to types.go.

controller.go
ser-io: done, used `apiv1` though

controller.go
ser-io: decided to keep controller.go internal.

jemoreira (approved):

PR #9, opened 2022-03-31
Improve errors returned to the clients and account for the v1 url prefix everywhere

jemoreira:
Errors returned to the client should be in the format expected by signaling server clients
Allow to specify different error messages for returning to client and logging
Account for the v1 prefix in more places in the signaling server (it was broken after the endpoints moved)
Don't require the deviceId query parameter.


controller.go
```diff
@@ -82,9 +82,10 @@ func (fn HTTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		status := http.StatusInternalServerError
 		var e *AppError
 		if errors.As(err, &e) {
-			status = e.StatusCode
+			replyJSON(w, e.JSONResponse(), e.StatusCode)
+		} else {
+			http.Error(w, err.Error(), status)
```
ser-io, line 88: this line could leak internal information as we do not enforce lower layers to return AppError.

Databean (approved):

controller.go
jemoreira: done

controller.go
```diff
@@ -80,11 +81,17 @@ func (fn HTTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if err := fn(w, r); err != nil {
 		log.Println("Error: ", err)
 		status := http.StatusInternalServerError
+		var response apiv1.ErrorMsg
```
ser-io, line 84: rather than using local vars like status and response to be modified down below, can we have something like?

if errors.As(err, &e) {
  replyJSON(w, e.JSONResponse(), e.StatusCode)
} else {
  replyJSON(w, apiv1.ErrorMsg{Error: "Internal Server Error"}, http.StatusInternalServerError)
}

I think it's easier to follow.

jemoreira: done

ser-io (approved):

PR #10, opened 2022-04-08
Splits main package.

ser-io:
Moves code into internal packages to reduce the public API surface.


jemoreira (commented):
I agree that we should organize better now that the number of files is growing, but I think this organization can be improved a little more:
- Having a directory named gcp that only contains a file named gcp.go and test_gcp.go seems a bit redundant to me.
- Why does the instance manager get its own package, but other modules like the signaling server and the account manager don't?

IMHO we shouldn't separate the IM from the AM and the SS, instead we should separate the GCP implementations from the non-GCP ones. The resulting hierarchy could look something like this:
app
| gcp
|  | instancemanager.go
|  | test_instancemanager.go
|  | gaeusersaccountmanager.go
| os
|  | osaccountmanager.go
| config.go
| errors.go
| ...

WDYT?

ser-io:
Done. Yes, I'm onboard with the suggested structure, less packages.

jemoreira (approved):

PR #11, opened 2022-04-25
Update Signaling Server routes with `zone` and `host` parameters.

ser-io:


app/controller.go
```diff
@@ -93,19 +93,23 @@ func (fn HTTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 }
 
 func (c *Controller) GetDeviceFiles(w http.ResponseWriter, r *http.Request, user UserInfo) error {
+	zone := mux.Vars(r)["zone"]
+	host := mux.Vars(r)["host"]
```
jemoreira, line 97: It's not a good idea to repeat these two lines too many times, especially since "zone" and "host" are used as literals. Could you encapsulate extracting the host and zone parameters from the request into a utility function?

jemoreira (changes requested):

app/controller.go
ser-io: Done, created to independent utility functions.

PR #12, opened 2022-04-27
Retrieves GCP host instance address.

ser-io:


app/types.go
```diff
@@ -44,22 +44,20 @@ type SignalingServer interface {
...
+	GetHostAddr(zone string, host string, user UserInfo) (string, error)
+	// Creates a host instance.
 	CreateHost(zone string, req *apiv1.CreateHostRequest, user UserInfo) (*apiv1.Operation, error)
 	// Closes the connection with the underlying API
 	Close() error
 }
 
+var ErrBadCreateHostRequest = NewBadRequestError("invalid CreateHostRequest", nil)
+var ErrHostInstanceNotOwnedByUser = NewForbiddenError("host instance not owned by user", nil)
+var ErrMissingNetworkInterface = NewInternalError("host instance missing a network interface", nil)
```
jemoreira, line 59: what stops code from another package from doing
```
gcp.ErrBadCreateHostRequest = <something else>
```
??

IIUC the only reason you need these variables is so you can test with `errors.Is()`, but the same effect can be achieved by making these types instead of variables and checking with `errors.As()`.

jemoreira (changes requested):

jemoreira (approved):

PR #13, opened 2022-05-05
Add file based configuration

jemoreira:


jemoreira:
Sorry, this isn't ready yet

jemoreira:
Now it's ready

app/config.go
```diff
@@ -14,20 +14,54 @@
...
+	toml "github.com/pelletier/go-toml/v2"
+	"os"
+)
+
+const DefaultConfFile = "conf.toml"
+const ConfFileEnvVar = "CONFIG_FILE"
+
 type Config struct {
-	GCPConfig *GCPConfig
+	AccountManager  AMConfig
```
ser-io, line 26: nit: should we use pointers here as well?

app/config.go
```diff
@@ -14,20 +14,54 @@
...
+	InstanceManager IMConfig
+	Infra           *InfraConfig
+}
+
+type IMConfig struct {
+	GCP *GCPIMConfig
+}
+type GCPIMConfig struct {
```
ser-io, line 34: nit: leave space above, I think `go fmt ./...` handles this.

app/config.go
jemoreira: Actually I prefer all of these members to be non-pointers since they are not optional (can't be nil). The GCPConfig field is a pointer because it would be nil if another implementation is ever used. We also don't transfer ownership of these member variables, they are only borrowed by the modules.

ser-io (approved):

PR #14, opened 2022-05-06
Check for correct formatting on new PRs

jemoreira:


.github/workflows/test_build_deb.yaml
```diff
@@ -16,6 +16,8 @@ jobs:
       uses: actions/checkout@a81bbbf8298c0fa03ea29cdc473d45769f953675 # aka v2
       with:
         path: "cloud-android-orchestration"
+    - name: check format
+      run: if [ "$(gofmt -s -l . | wc -l)" -gt 0 ]; then exit 1; fi
```
Databean, line 20: nit: Could this be a separate file under `.github/workflows/` to validate the formatting and other tests separately?

jemoreira: Done
Sorry for the force-push spam, I was trying to reuse the stuff repeated between the workflows, but it seems it's not possible or straighforward.

Databean (approved):

PR #16, opened 2022-05-19
Fix PlaceholderIM and make it choosable via configuration

jemoreira:


conf.toml
```diff
@@ -2,6 +2,7 @@
 Type = "unix"
 
 [InstanceManager.GCP]
+Type = "unix"
```
ser-io, line 5: shouldn't this be at an upper level?

app/config.go
```diff
@@ -29,9 +29,17 @@ type Config struct {
...
-	GCP *GCPIMConfig
+	Type IMType
+	GCP  *GCPIMConfig
 }
 
+type IMType string
+
+const (
+	UnixIMType IMType = "unix"
+	GCEIMType  IMType = "GCE"
```
ser-io, line 40: should we use "GCP" to keep it consistent?

main.go
```diff
@@ -29,9 +29,16 @@ func main() {
...
+	var im app.InstanceManager
+	if config.InstanceManager.Type == app.GCEIMType {
+		im, err = gcp.NewInstanceManager(&config.InstanceManager, context.Background())
+		if err != nil {
+			log.Fatal(err)
+		}
+	} else if config.InstanceManager.Type == app.UnixIMType {
+		im = &unix.InstanceManager{}
+	} else {
+		log.Fatal("Unknown Instance Manager type: ", config.InstanceManager.Type)
```
ser-io, line 41: nit: consider a switch statement.

app/unix/instancemanager.go
```diff
@@ -12,23 +12,24 @@
...
 
-func (m *PlaceholderIM) GetHostAddr(_ string, _ string, _ UserInfo) (string, error) {
+func (d *InstanceManager) GetHostAddr(_ string, _ string) (string, error) {
 	return "127.0.0.1", nil
 }
 
-func (d *PlaceholderIM) CreateHost(_ string, _ *apiv1.CreateHostRequest, _ UserInfo) (*apiv1.Operation, error) {
-	return nil, NewInternalError("PlaceholderIM::CreateHost is not implemented", nil)
+func (d *InstanceManager) CreateHost(_ string, _ *apiv1.CreateHostRequest, _ app.UserInfo) (*apiv1.Operation, error) {
+	return nil, app.NewInternalError("unix.InstanceManager::CreateHost is not implemented", nil)
```
ser-io, line 32: nit: wondering whether using a C++ nomenclature could be avoided? I think `CreateHost method is not implemented` is fine.

conf.toml
jemoreira: right, good catch!

app/config.go
jemoreira: done, changed the account manager type too

main.go
jemoreira: done

app/unix/instancemanager.go
jemoreira: I want to include the class name because this will printed without context somewhere else, changed to class#method as it's less language specific

ser-io: sounds good, suggest to use  "%T" format verb then to avoid hard-coding the type.

ser-io (approved):

app/unix/instancemanager.go
jemoreira: done

PR #22, opened 2022-06-29
List hosts

ser-io:


app/controller.go
```diff
@@ -157,6 +158,36 @@ func (c *Controller) CreateHost(w http.ResponseWriter, r *http.Request, user Use
...
+		return err
+	}
+	replyJSON(w, res, http.StatusOK)
+	return nil
+}
+
+func BuildListHostRequest(r *http.Request) (*ListHostsRequest, error) {
+	var maxResults int
+	maxResultsRaw := r.URL.Query().Get("maxResults")
+	if maxResultsRaw != "" {
```
jemoreira, line 177: might be worth it to encapsulate this in a utility function to parse int query parameters

app/gcp/instancemanager.go
```diff
@@ -104,6 +104,39 @@ func (m *InstanceManager) CreateHost(zone string, req *apiv1.CreateHostRequest,
 	return result, nil
 }
 
+const listHostsRequestMaxResultsLimit = 500
+
+func (m *InstanceManager) ListHosts(zone string, user app.UserInfo, req *app.ListHostsRequest) (*apiv1.ListHostsResponse, error) {
+	if req.MaxResults > listHostsRequestMaxResultsLimit {
```
jemoreira, line 110: this doesn't have to be an error, you can return at most listHostsRequestMaxResultsLimit elements and a nextPageToken instead.

jemoreira (changes requested):

app/controller.go
ser-io: done

app/gcp/instancemanager.go
ser-io: done

jemoreira (approved):

PR #23, opened 2022-08-16
Forward relevant requests to host orchestrator.

ser-io:


app/controller.go
```diff
@@ -89,6 +98,53 @@ func (fn HTTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		if zone == "" {
+			return NewBadRequestError("Invalid url, missing zone value", nil)
+		}
+		host := vars["host"]
+		if host == "" {
+			return NewBadRequestError("Invalid url, missing host value", nil)
+		}
+		split := strings.SplitN(r.URL.String(), "hosts/"+host+"/", 2)
+		if len(split) != 2 {
+			return NewBadRequestError("Invalid url, missing host resource", nil)
```
jemoreira, line 121: Is this possible? The matcher specified in the route should guarantee that the URL contains a non empty prefix before the separator string and a non empty suffix after it. Which means this should be an assertion turning into a server error (missconfigured server routes) when found instead of a bad request from the user.

app/controller.go
```diff
@@ -89,6 +98,53 @@ func (fn HTTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		req, err := http.NewRequest(r.Method, url, r.Body)
+		if err != nil {
+			return nil
+		}
+		resp, err := f.Client.Do(req)
+		if err != nil {
+			return err
+		}
+		defer resp.Body.Close()
+		b, err := ioutil.ReadAll(resp.Body)
```
jemoreira, line 137: You could avoid most of this boilerplate by using an instance of [ReverseProxy](https://pkg.go.dev/net/http/httputil#ReverseProxy). I think you'll only need to provide an implementation for `Director` and leave everything else as the default. Or you could use `NewSingleHostReverseProxy` as is done in `signalingserver.go`.

jemoreira (changes requested):

app/controller.go
ser-io: good catch, totally agree. done!

app/controller.go
```diff
@@ -98,53 +96,41 @@ func (h HTTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-		defer resp.Body.Close()
-		b, err := ioutil.ReadAll(resp.Body)
+		hostURL, err := f.URLResolver.GetHostURL(zone, host)
 		if err != nil {
 			return err
 		}
-		w.Header().Set(headerContentType, resp.Header.Get(headerContentType))
-		w.WriteHeader(resp.StatusCode)
-		w.Write(b)
+		split := strings.SplitN(r.URL.Path, "hosts/"+host, 2)
```
jemoreira, line 118: I think the `Director` function is the most natural place to do these URL transformations: It's purpose is to transform one URL into the other one after all

app/controller.go
ser-io: definitely better, created the `ReverseProxy` instance passing the `Director` to avoid modifying the path of the original request.

app/controller.go
ser-io: acknowledged, done.

jemoreira: I'm sorry I wasn't clear enough in my comment. My suggestion is to create a proxy object that could be re-used between requests. In the `Director` function it will extract variables from the request (with `mux.Var()`) and figure out the hostURL.

ser-io: I liked the idea of having a single proxy instance and managing the routing logic into the Director function, however didn't find it feasible due the Director function does not expect error in its logic, and fetching the host url could result in errors, hence I decided to create reverse proxy instances only when the request is valid, created a hostReverseProxy type though in case we want to cache instances later on.

jemoreira: you could have an error returning logic from the `Director` logic by simply changing the url to an error page, but that's not worth it just to have a single proxy object

jemoreira (approved):

PR #25, opened 2022-09-14
Revert "Use acloud ownership label. (#24)"

ser-io:
This reverts commit de5bbbbfe4d7ba197bcf5bd80b6ef37a41ea29ff.

- Not the acloud ownership label.


jemoreira:
This doesn't look like a complete revert of that PR, maybe a different name should be better

jemoreira (approved):

PR #28, opened 2022-09-21
Adds extra step to avoid creating instances in projects used by acloud.

ser-io:
- Creating new host and cvd instances in projects used by acloud could corrupt current acloud functionalities. It's safe to use Cloud Orchestration for "Reading" operations like webrtc streaming.

### DIFF
--- a/app/config.go
+++ b/app/config.go
@@ -26,6 +26,7 @@ type Config struct {
 	AccountManager  AMConfig
 	InstanceManager IMConfig
 	Infra           InfraConfig
+	Operations      OperationsConfig
 }
 
 type IMConfig struct {
@@ -41,9 +42,8 @@ const (
 )
 
 type GCPIMConfig struct {
-	ProjectID               string
-	HostImage               string
-	ContainsAcloudInstances bool
+	ProjectID string
+	HostImage string
 }
 
 type AMConfig struct {
@@ -59,6 +59,10 @@ const (
 
 type InfraConfig struct {
 	STUNServers []string
+}
+
+type OperationsConfig struct {
+	CreateHostDisabled bool
 }
 
 func LoadConfig() (*Config, error) {
--- a/app/config.go
+++ b/app/config.go
@@ -41,8 +41,9 @@ const (
 )
 
 type GCPIMConfig struct {
-	ProjectID string
-	HostImage string
+	ProjectID               string
+	HostImage               string
+	ContainsAcloudInstances bool
 }
 
 type AMConfig struct {
--- a/app/controller.go
+++ b/app/controller.go
@@ -35,16 +35,20 @@ import (
 // relevant modules
 type Controller struct {
 	infraConfig     apiv1.InfraConfig
+	opsConfig       OperationsConfig
 	instanceManager InstanceManager
 	sigServer       SignalingServer
 	accountManager  AccountManager
 }
 
-func NewController(servers []string, im InstanceManager, ss SignalingServer, am AccountManager) *Controller {
-
+func NewController(
+	servers []string,
+	opsConfig OperationsConfig,
+	im InstanceManager,
+	ss SignalingServer,
+	am AccountManager) *Controller {
 	infraCfg := buildInfraCfg(servers)
-	controller := &Controller{infraCfg, im, ss, am}
-	return controller
+	return &Controller{infraCfg, opsConfig, im, ss, am}
 }
 
 func (c *Controller) Handler() http.Handler {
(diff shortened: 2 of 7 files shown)


app/config.go
```diff
@@ -41,8 +41,9 @@ const (
 )
 
 type GCPIMConfig struct {
-	ProjectID string
-	HostImage string
+	ProjectID               string
+	HostImage               string
+	ContainsAcloudInstances bool
```
jemoreira, line 46: This shouldn't mention "acloud". Create a config entry that simply disables creating and destroying hosts. This could make sense in projects where acloud is not in use, for example if a user's host are created by a different entity and shared with them read-only.

app/gcp/instancemanager.go
```diff
@@ -71,6 +71,9 @@ func (m *InstanceManager) GetHostURL(zone string, host string) (*url.URL, error)
 const operationStatusDone = "DONE"
 
 func (m *InstanceManager) CreateHost(zone string, req *apiv1.CreateHostRequest, user app.UserInfo) (*apiv1.Operation, error) {
+	if m.Config.GCP.ContainsAcloudInstances {
```
jemoreira, line 74: This should be in the controller instead, since it's disabling the entire endpoint.

app/gcp/instancemanager.go
```diff
@@ -71,6 +71,9 @@ func (m *InstanceManager) GetHostURL(zone string, host string) (*url.URL, error)
...
+		return nil, app.NewBadRequestError("Invalid CreateHostRequest: GCP project contains acloud instances", nil)
```
jemoreira, line 75: Http code 405 Method Not Allowed is more appropriate

jemoreira (changes requested):

app/config.go
ser-io: done.

app/gcp/instancemanager.go
ser-io: done

app/gcp/instancemanager.go
ser-io: done.

app/controller.go
```diff
@@ -193,6 +197,9 @@ func (c *Controller) forward(w http.ResponseWriter, r *http.Request, user UserIn
 }
 
 func (c *Controller) createHost(w http.ResponseWriter, r *http.Request, user UserInfo) error {
+	if c.opsConfig.CreateHostDisabled {
+		return NewMethodNotAllowedError("Create host operation is disabled", nil)
+	}
```
jemoreira, line 202: Given that this config value won't change while the service is running it might make more sense to check for it in the `Handler()` method. If it's disabled you can register a different handler that always returns 405. WDYT?

app/config.go
```diff
@@ -60,6 +61,10 @@ type InfraConfig struct {
 	STUNServers []string
 }
 
+type OperationsConfig struct {
+	CreateHostDisabled bool
```
jemoreira, line 65: Use `CreateHostEnabled` instead.

I think I've seen a recommendation not to use negative (disabled) in config properties, but I couldn't find any official guidance. The idea is that using "Disabled" makes it confusing: for example, the docs could say "When `CreateHostDisabled` is disabled Host Creation is enabled".

app/controller.go
ser-io: good point, done!

app/config.go
ser-io: I see this relevant when the config parameter always needs to be set, in this case it could be omitted which makes a better default IMO vs having to set `CreateHostEnabled = true` every time. Using the omitting way docs should only point out that if you want to disable host creation add `CreateHostDisabled = true`.

app/controller.go
```diff
@@ -81,6 +81,13 @@ func (c *Controller) Handler() http.Handler {
 	return router
 }
 
+func (c *Controller) createHostHTTPHandler() HTTPHandler {
+	if c.opsConfig.CreateHostDisabled {
+		return notAllowedHttpHandler()
```
jemoreira, line 86: nit: Instead of `notAllowedHttpHandler` being a function that creates and returns an http handler it could be the handler itself and be returned here:
```
return notAllowedHttpHandler // without the () at the end

...

func notAllowedHttpHandler(w http.ResponseWriter, r *http.Request) error {
    return NewMethodNotAllowedError("Operation is disabled", nil)
}
```

jemoreira (approved):

app/controller.go
ser-io: yes indeed, good catch! thanks

PR #37, opened 2022-10-14
Delete host instances.

ser-io:


app/gcp/instancemanager.go
```diff
@@ -148,7 +148,27 @@ func (m *InstanceManager) ListHosts(zone string, user app.UserInfo, req *app.Lis
 }
 
 func (m *InstanceManager) DeleteHost(zone string, user app.UserInfo, name string) (*apiv1.Operation, error) {
-	return nil, app.NewInternalError(fmt.Sprintf("%T#DeleteHost is not implemented", *m), nil)
+	nameFilterExpr := "name=" + name
+	ownerFilterExpr := fmt.Sprintf("labels.%s:%s", labelAcloudCreatedBy, user.Username())
+	res, err := m.Service.Instances.
+		List(m.Config.GCP.ProjectID, zone).
+		Context(context.TODO()).
+		Filter(fmt.Sprintf("%s AND %s", nameFilterExpr, ownerFilterExpr)).
```
jemoreira, line 156: Is it possible to merge the list and delete api calls into one? IOW, can you use `Filter()` in the delete call and skip the `List()` call altogether?

ser-io: Unfortunately not, the delete operation does not offer a filtering logic. Double checked the list of api methods https://cloud.google.com/compute/docs/reference/rest/v1/instances/ looking for a method offering both operations but found none.

jemoreira (approved):

PR #39, opened 2022-10-27
Restructures project layout to allow multiple binaries.

ser-io:


cmd/server/main.go
```diff
@@ -20,9 +20,9 @@ import (
 	"net/http"
```
jemoreira, line 20: "server" is possibly too generic. What about "cloud_orchestrator"?

ser-io: true, changed it to "cloud_orchestrator"

jemoreira (approved):

PR #42, opened 2022-10-31
[cvdremote] Adds sub commands routing logic.

ser-io:


internal/cvdremote/cvdremote.go
```diff
@@ -0,0 +1,131 @@
...
+
+func (c *CVDRemoteCommand) Run(args []string) error {
+	if err := c.fs.Parse(args); err != nil {
+		return err
+	}
+	args = c.fs.Args()
+	if len(args) == 0 {
+		return fmt.Errorf("missing resource")
+	}
+	var resource string
```
jemoreira, line 109: nit:
```
// If no resource is given it defaults to "cvd".
// There is no chance of a collision between a cvd subcommand and a resource name because
// resources are nouns and subcommands are verbs.
resource = resourceCVD
if _,ok := c.subCommandsMap[args[0]]; ok {
    resource = args[0]
    args = args[1:]
}
```

internal/cvdremote/cvdremote.go
```diff
@@ -0,0 +1,131 @@
...
+
+    list
+        List the resources.`
+
+type CVDRemoteCommand struct {
+	fs             *flag.FlagSet
+	subCommandsMap map[string]map[string]Command
+}
+
+func NewDefaultCVDRemoteCommand() Command {
```
jemoreira, line 60: nit: The use of "Default" in the name implies that there are other valid `CVDRemoteCommand`s, which  is only true during testing. This function should be called `NewCVDRemoteCommand` and for testing you should call one called `NewTestingCVDRemoteCommand`.

internal/cvdremote/cvdremote.go
```diff
@@ -0,0 +1,131 @@
...
+type Command interface {
+	Run([]string) error
+}
+
+// Groups the subcommands handled in cvdremote.
+type CVDRemoteSubCommands struct {
+	CreateCVD   Command
+	ListCVDs    Command
+	ListHosts   Command
+	CreateHost  Command
```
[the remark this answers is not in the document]
ser-io: done

internal/cvdremote/cvdremote.go
```diff
@@ -0,0 +1,131 @@
...
+	DeleteHosts Command
```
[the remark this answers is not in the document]
ser-io: adding these in bulk as these are the ones implemented already in the aosp version, new sub-commands could be added later one by one.

internal/cvdremote/cvdremote.go
ser-io: done

internal/cvdremote/cvdremote.go
ser-io: used `NewCVDRemoteCommand` and `NewCVDRemoteCommandWithArgs`

jemoreira (approved):

PR #45, opened 2022-11-04
[cli] Implement create host functionality.

ser-io:
- Introduces the Cobra library (https://github.com/spf13/cobra).


pkg/cli/cli.go
```diff
@@ -0,0 +1,207 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runner.Run()
+		},
+	}
+	cmd.Flags().StringVar(&machineType, machineTypeFlag, "n1-standard-4", "Machine type.")
+	cmd.Flags().StringVar(&minCPUPlatform, minCpuPlatformFlag, "Intel Haswell",
+		"Specifies a minimum CPU platform for the VM instance.")
+	runner.MachineType = &machineType
+	runner.MinCPUPlatform = &minCPUPlatform
+	runner.Command = cmd
```
jemoreira, line 128: If you pass the `cmd` as a parameter to `runner.Run()` you won't need to hold a reference to it from the runner struct, which avoids the circular dependency between runner and command.

Without that circular dependency you can instantiate the entire runner before the command.

Also, make `MachineType` and `MinCPUPlatform` strings instead of pointers, with the command taking its addresses in the flag specification.

jemoreira (changes requested):

pkg/cli/cli.go
ser-io: good catch! done as you mentioned. removed the handling of `MachineType` and `MinCPUPlatform` as these are not yet handled

internal/cvdremote/cvdremote.go
```diff
@@ -1,131 +0,0 @@
-// Copyright 2022 Google LLC
```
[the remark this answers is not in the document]
ser-io: yes, full delete.

pkg/cli/cli.go
```diff
@@ -0,0 +1,187 @@
...
+		err := &apiCallError{op.Result.Error}
+		return err
+	}
+	if !op.Done {
+		return opTimeoutError(op.Name)
+	}
+	ins := new(apiv1.HostInstance)
+	if err := json.Unmarshal([]byte(op.Result.Response), ins); err != nil {
+		return err
+	}
```
jemoreira, line 149: nit: It's not wrong, but there is no need for `ins` to be a pointer (it's not shared and can't be `nil`). I'd prefer for `ins` to be an object and pass its address to `Unmarshal`

pkg/cli/cli.go
```diff
@@ -0,0 +1,187 @@
...
+	req, err := http.NewRequest(method, url, body)
+	req.Header.Set("Content-Type", "application/json")
+	res, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
+	dec := json.NewDecoder(res.Body)
+	if res.StatusCode < 200 || res.StatusCode > 299 {
+		errPayload := new(apiv1.Error)
```
jemoreira, line 175: nit: same comment about errPayload not having to be a pointer

pkg/cli/cli.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	return cmd
+}
+
+func runCreateHostCommand(c *cobra.Command) error {
+	serviceURL := c.InheritedFlags().Lookup(serviceURLFlag).Value.String()
+	baseURL := serviceURL + "/v1"
+	req := &apiv1.CreateHostInstanceRequest{}
+	body := &apiv1.CreateHostRequest{CreateHostInstanceRequest: req}
+	op := new(apiv1.Operation)
```
jemoreira, line 131: ni: same comment as below about `op` not having to be a pointer.

jemoreira (approved):

pkg/cli/cli.go

pkg/cli/cli.go

pkg/cli/cli.go
ser-io: kept this one as a pointer as it's returned.

PR #48, opened 2022-11-08
Allows http communication through proxy using flag --http_proxy

ser-io:


pkg/cli/cli.go
```diff
@@ -134,16 +146,32 @@ func newHostCommand() *cobra.Command {
...
-func runCreateHostCommand(c *cobra.Command) error {
-	baseURL := buildBaseURL(c)
+func runSubCommand(c *cobra.Command, args []string, runner subCommandRunner) error {
+	proxyURL := c.InheritedFlags().Lookup(httpProxyFlag).Value.String()
+	// Handles http proxy
+	if proxyURL != "" {
+		if _, err := url.Parse(proxyURL); err != nil {
+			return err
+		}
+		os.Setenv("HTTP_PROXY", proxyURL)
```
jemoreira, line 156: Any reason not to use https://pkg.go.dev/net/http#Transport to set the proxy?
An environment variable is no different than a global variable.

ser-io: good catch! used Transport.

jemoreira (approved):

PR #49, opened 2022-11-09
[cli] Implements delete hosts operation

ser-io:


pkg/cli/cli.go
```diff
@@ -235,6 +263,9 @@ func doRequest(client *http.Client, method, url string, reqPayload interface{},
 	defer res.Body.Close()
 	dec := json.NewDecoder(res.Body)
 	if res.StatusCode < 200 || res.StatusCode > 299 {
+		if method == "DELETE" {
+			return &apiCallError{&apiv1.Error{Code: strconv.Itoa(res.StatusCode)}}
```
jemoreira, line 267: Would it make sense to use `res.Status` instead of `strconv.Itoa(res.StatusCode)`. The status will have more information than the code, the downside is that it won't be easy to parse back into an int later.

pkg/cli/cli.go
```diff
@@ -235,6 +263,9 @@ func doRequest(client *http.Client, method, url string, reqPayload interface{},
 	defer res.Body.Close()
 	dec := json.NewDecoder(res.Body)
 	if res.StatusCode < 200 || res.StatusCode > 299 {
+		if method == "DELETE" {
```
jemoreira, line 266: Add a comment explaining why `DELETE` methods need different error handling.

pkg/cli/cli.go
```diff
@@ -202,6 +213,27 @@ func runListHostsCommand(c *cobra.Command, _ []string, opts *subCommandOptions)
...
+	var errs error
+	for _, arg := range args {
+		wg.Add(1)
+		go func(name string) {
+			defer wg.Done()
+			url := opts.BaseURL + "/hosts/" + name
+			if err := doRequest(opts.HTTPClient, "DELETE", url, nil, nil); err != nil {
+				mu.Lock()
+				errs = multierror.Append(errs, fmt.Errorf("delete host %q failed: %w", name, err))
+				defer mu.Unlock()
```
jemoreira, line 228: Either move this line to right after the call to `Lock()` or don't use defer (if you know nothing will return in between lock and unlock). By the time control reaches this line there is no need to defer anymore

pkg/cli/cli.go
ser-io: done. Good point. Uses the message property instead `res.Status`

pkg/cli/cli.go
ser-io: good catch! done

pkg/cli/cli.go
ser-io: done

jemoreira (approved):

PR #51, opened 2022-11-10
Add webrtc client library

jemoreira:
For the time being it's only capable of communicating over the ADB channel.


pkg/webrtcclient/client.go
```diff
@@ -0,0 +1,234 @@
...
+	ClientICEServers []webrtc.ICEServer
+}
+
+type Controller struct {
+	signaling      *Signaling
+	peerConnection *webrtc.PeerConnection
+	observer       Observer
+}
+
+func (c Controller) connect(onComplete func(error)) {
```
ser-io, line 40: is the value receiver intended here? If so, lets add a comment.

pkg/webrtcclient/client.go
```diff
@@ -0,0 +1,234 @@
...
+			}
+			switch msg["type"] {
+			case RequestOfferMsgType:
+				panic("Device requested an offer. This violates the signaling protocol")
+			case OfferMsgType:
+				var offer webrtc.SessionDescription
+				err := Reshape(msg, &offer)
+				if err != nil {
+					// msg was obtained from a JSON string so Reshape shouldn't fail
+					panic(fmt.Errorf("Failed to reshape json: ", err))
```
ser-io, line 97: shouldn't we add a format verb here, i.e `%v`?

ser-io, line 97: nit: if this is just for printing the panic message why not use fmt.Sprintf instead of creating a new error.

pkg/webrtcclient/client.go
```diff
@@ -0,0 +1,234 @@
...
+		case msg, open := <-c.signaling.RecvCh:
+			if !open {
+				return fmt.Errorf("Signaling channel closed")
+			}
+			switch msg["type"] {
+			case RequestOfferMsgType:
+				panic("Device requested an offer. This violates the signaling protocol")
+			case OfferMsgType:
+				var offer webrtc.SessionDescription
+				err := Reshape(msg, &offer)
```
ser-io, line 94: nit:if with init statement could be applied here and in similar cases down below

for example:
`
if err := Reshape(msg, &offer); err != nil {
  // msg was obtained from a JSON string so Reshape shouldn't fail
  panic(fmt.Errorf("Failed to reshape json: ", err))
}
`

pkg/webrtcclient/client.go
```diff
@@ -0,0 +1,234 @@
...
+				if err != nil {
+					// msg was obtained from a JSON string so Reshape shouldn't fail
+					panic(fmt.Errorf("Failed to reshape json: ", err))
+				}
+				err = c.onOffer(offer)
+				if err != nil {
+					return fmt.Errorf("Error handling offer: %v", err)
```
ser-io, line 101: nit: %w if you want to embed `err` into the one created.

pkg/webrtcclient/types.go
```diff
@@ -0,0 +1,70 @@
...
+		candidate,
+	}
+}
+
+// The offer and answer messages have the same structure as the
+// webrtc.SessionDescription so they don't to be redefined here.
+
+// Allows converting a generic JSON object representation, like
+// map[string]interface{}, to a specific class such as ICECandidateMsg.
+func Reshape(original interface{}, reshaped interface{}) error {
```
ser-io, line 60: nit: Go 1.18 introduced generics which could make this function easier to read.

pkg/webrtcclient/client.go
jemoreira: nope, my mistake

pkg/webrtcclient/client_test.go
```diff
@@ -0,0 +1,153 @@
...
+	var msg webrtc.SessionDescription
+	if err := recvWithTimeOut(signaling.SendCh, &msg); err != nil {
+		t.Fatalf("Client didn't send the answer message in time: %v", err)
+	}
+	if msg.Type != webrtc.SDPTypeAnswer {
+		t.Fatalf("Expected 'answer' message, but got '%v' instead", msg.Type)
+	}
+}
+
+func TestClientGeneratesICECandidate(t *testing.T) {
```
ser-io, line 123: nit: I find it useful to lay out the test methods into 3 clear blocks: arrange, act, and assert separated by new line.

pkg/webrtcclient/types.go
ser-io: I was thinking on avoiding the second parameter and returning: `(*K, error) instead. 

On a different note, can we specify original as map[string]interface{}?

pkg/webrtcclient/client_test.go
jemoreira: Done

ser-io (approved):

pkg/webrtcclient/types.go
jemoreira: Refactored to use generics.

Can't make original be `map` because it's used on objects received from the server, which could be arrays too.

PR #62, opened 2022-11-19
adbtunnel command

jemoreira:


pkg/cli/adbtunnel.go
```diff
@@ -0,0 +1,231 @@
...
+// limitations under the License.
+
+package cli
+
+import (
+	"fmt"
+	"net"
+	"sync/atomic"
+
+	wclient "github.com/google/cloud-android-orchestration/pkg/webrtcclient"
```
ser-io, line 22: nit: please add a new line down below to separate the internal and third party import blocks

pkg/cli/adbtunnel.go
```diff
@@ -0,0 +1,231 @@
...
+	"github.com/hashicorp/go-multierror"
+	"github.com/pion/webrtc/v3"
+	"github.com/spf13/cobra"
+)
+
+const hostFlag string = "host"
```
ser-io, line 28: nit: could this be moved to cli.go, it's going to be used by `cvd` subcommand as well?

pkg/cli/adbtunnel.go
jemoreira: done

pkg/cli/adbtunnel.go
jemoreira: done

pkg/cli/adbtunnel.go
```diff
@@ -0,0 +1,234 @@
...
+type ADBForwarder struct {
+	cmd      *command
+	dc       *webrtc.DataChannel
+	host     string
+	device   string
+	listener net.Listener
+	conn     net.Conn
+	port     int
+	running  atomic.Bool
+}
```
ser-io, line 111: nit: wondering whether it's better to isolate the immutable properties like `cmd`, `host` and `device` into a sub structure, so it's more clear what changes and what not.

ser-io (approved):

pkg/cli/adbtunnel.go
jemoreira: I don't like having those 3 in the ADBForwarder either, the only reason is so I can print logs. But I don't think the extra structure will help readability.

Grouped them together and added an empty line to separate them instead.

PR #63, opened 2022-11-19
[orchestrator] Change wait operation response.

ser-io:
- Return the direct result or error rather than the whole operation, having the responses for failed operations be aligned with the relevant http status code.
- Remove result property from operation object to avoid the result being queried this way.

### DIFF
--- a/api/v1/instancemanager.go
+++ b/api/v1/instancemanager.go
@@ -32,7 +32,7 @@ type Operation struct {
 	// available.
 	Done bool `json:"done"`
 	// Result will contain either an error or a result object but never both.
-	Result *OperationResult `json:"result,omitempty"`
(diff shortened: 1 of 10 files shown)


api/v1/instancemanager.go
```diff
@@ -32,7 +32,7 @@ type Operation struct {
 	// available.
 	Done bool `json:"done"`
 	// Result will contain either an error or a result object but never both.
-	Result *OperationResult `json:"result,omitempty"`
+	// Result *OperationResult `json:"result,omitempty"`
```
jemoreira, line 35: this should be deleted

jemoreira (approved):

api/v1/instancemanager.go
ser-io: done

PR #74, opened 2022-12-03
[api] Use a single error object

ser-io:


api/v1/signalingserver.go
```diff
@@ -1,7 +1,8 @@
 package v1
 
-type ErrorMsg struct {
-	Error string `json:"error"`
+type Error struct {
+	Code     string `json:"code"`
+	ErrorMsg string `json:"error"`
```
jemoreira, line 5: FYI: This is ok only because you kept the `error` string field. This message is sent to devices, which expect that field to be there.

jemoreira (approved):

api/v1/signalingserver.go
ser-io: true, yes otherwise it will be an api breaking change.

PR #76, opened 2022-12-05
[cli] Do not set a default value for gcp_min_cpu_platform flag

ser-io:


pkg/cli/host.go
```diff
@@ -50,7 +50,7 @@ func newHostCommand(cfgFlags *configFlags) *cobra.Command {
 	}
 	create.Flags().StringVar(&createFlags.MachineType, gcpMachineTypeFlag, "n1-standard-4",
 		"Indicates the machine type")
-	create.Flags().StringVar(&createFlags.MinCPUPlatform, gcpMinCPUPlatformFlag, "Intel Haswell",
```
jemoreira, line 53: I assume that this not configured in the gce api request if the value is not set, correct?

jemoreira (approved):

pkg/cli/host.go
ser-io: correct, if set as empty it's handled as not set.

PR #79, opened 2022-12-06
Fix value of Min CPU platform to "Intel Haswell".

ser-io:
- "Intel Haswell" is required as min cpu platform in order to enable nested virtualization. https://cloud.google.com/compute/docs/instances/nested-virtualization/enabling
- Remove the ability to set this parameter from clients.

### DIFF
--- a/pkg/app/gcp/instancemanager.go
+++ b/pkg/app/gcp/instancemanager.go
@@ -85,8 +85,7 @@ func (m *InstanceManager) CreateHost(zone string, req *apiv1.CreateHostRequest, 
 		Name: m.InstanceNameGenerator.NewName(),
 		// This is required in the format: "zones/zone/machineTypes/machine-type".
 		// Read more: https://cloud.google.com/compute/docs/reference/rest/v1/instances/insert#request-body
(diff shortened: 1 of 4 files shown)


jemoreira (approved):

pkg/app/gcp/instancemanager.go
```diff
@@ -107,6 +106,12 @@ func (m *InstanceManager) CreateHost(zone string, req *apiv1.CreateHostRequest,
 			},
 		},
 		Labels: labels,
+		// Required to enable nested virtualization
+		// https://cloud.google.com/compute/docs/instances/nested-virtualization/enabling#enabling_nested_virtualization_directly_on_a_new_vm
+		MinCpuPlatform: "Intel Haswell",
```
tardyp, line 111: This setting is actually only required for n1-* types instances.
It will not work for n2-* instances even if nested virtualization is supported.

So you actually need to add this only is MachineType startswith "n1-"

We found that when working will cloud workstation team to enable cuttlefish on this infra. The documentation is not very clear about that.

PR #80, opened 2022-12-06
[orchestrator] Uses GCP default cpu platform.

ser-io:


jemoreira (changes requested):
Could this be "uses default CPU platform unless the user specifies one"?

PR #85, opened 2023-01-06
Make cvdremote client configurable

jemoreira:
The `cvdremote` client can be configure with a file in toml format.
The file location is passed to the executable as an environment variable to allow for different locations depending on platform. Final users are not expected to pass this variable to the command, instead when the binary is distributed it should be wrapped in a way that ensures the variable is set with the appropriate path (for example, in debian `cvdremote` should point to a script that sets the variable and executes the binary)

### DIFF
--- a/pkg/cli/host.go
+++ b/pkg/cli/host.go
@@ -35,7 +35,7 @@ type createGCPHostFlags struct {
 	MinCPUPlatform string
 }
 
-func newHostCommand(cfgFlags *configFlags) *cobra.Command {
+func newHostCommand(cfgFlags *configFlags, config *HostConfig) *cobra.Command {
 	hostFlags := &subCommandFlags{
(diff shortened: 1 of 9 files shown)


pkg/cli/host.go
```diff
@@ -35,7 +35,16 @@ type createGCPHostFlags struct {
 	MinCPUPlatform string
 }
 
-func newHostCommand(cfgFlags *configFlags) *cobra.Command {
+type GCPHostConfig struct {
```
ser-io, line 38: could this be moved alongside the more general Config structure? having the config struct/parameters closer to each other makes it more easy to scan through all of them at once.

jemoreira: Done

ser-io (approved):

PR #86, opened 2023-01-07
Build cvdremote debian package

jemoreira:


build/package/debian/control
```diff
@@ -0,0 +1,17 @@
+Source: cvdremote
+Maintainer: Cuttlefish Team
+Section: misc
+Priority: optional
+Build-Depends: config-package-dev,
```
ser-io, line 5: are these required?

ser-io (approved):

build/package/debian/control
jemoreira: Don't know why I thought curl was `required`..., will delete

PR #88, opened 2023-01-10
[cli] Computes the list of required image files for local builds

ser-io:


pkg/cli/cvd.go
```diff
@@ -178,3 +231,28 @@ func (o *CVDOutput) String() string {
...
 	return res
 }
+
+const cvdHostPackageName = "cvd-host_package.tar.gz"
+
+func listRequiredImageFilenames(listFilename, androidProductOutDir, androidHostOutDir string) ([]string, error) {
+	var result []string
+	f, err := os.Open(listFilename)
+	if err != nil {
+		return nil, err
```
jemoreira, line 241: please wrap this error with some other string message, otherwise it's hard to debug when this happens because it will only say why can't the file be opened but won't have an indication of where in the code it's being opened. Something like:
```
return nil, fmt.Errorf("Error reading required file names: %w", err)
```

pkg/cli/cvd.go
```diff
@@ -178,3 +231,28 @@ func (o *CVDOutput) String() string {
...
+	}
+	defer f.Close()
+	r := bufio.NewReader(f)
+	for {
+		line, err := r.ReadString('\n')
+		line = strings.TrimRight(line, "\n")
```
jemoreira, line 247: touching `line` without checking `err` doesn't seem correct.

pkg/cli/cvd.go
```diff
@@ -178,3 +231,28 @@ func (o *CVDOutput) String() string {
...
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			return nil, err
+		}
+		result = append(result, androidProductOutDir+"/"+line)
+	}
+	result = append(result, androidHostOutDir+"/"+cvdHostPackageName)
```
jemoreira, line 256: The code above this line seems more complicated than necessary: The file you're reading is small enough that this should work:
```
content, err := os.ReadFile(listFilename)
if err != nil {
    // ...
}
result := strings.Split(string(content), "\n")
```

pkg/cli/cvd.go
```diff
@@ -178,3 +231,28 @@ func (o *CVDOutput) String() string {
...
+	defer f.Close()
+	r := bufio.NewReader(f)
+	for {
+		line, err := r.ReadString('\n')
+		line = strings.TrimRight(line, "\n")
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			return nil, err
```
jemoreira, line 252: same comment about wrapping the error

pkg/cli/cli_test.go
```diff
@@ -312,6 +312,44 @@ func TestDeleteHostsCommandFails(t *testing.T) {
 
 }
 
+func TestCreateCVDLocalImageMissingEnvVariables(t *testing.T) {
```
jemoreira, line 315: This test seems wrong to me:
1. It depends on an implementation detail: the fact that the env variables are checked before everything else instead of when needed.
2. If the code is modified to depend on a new variable this test won't be affected and there is no indication to the developer that they should add that new variable here.
3. It's not exactly incorrect to not read all 3 variables, but if you decide that 2 are enough then the test will fail even when the code is correct.

A more useful test would be to ensure the function doesn't fail when the 3 variables are set. You could create a temporary directory structure with the expected files and mock the `apiClient` to do nothing on upload to make a test like that work.

pkg/cli/cli_test.go
```diff
@@ -312,6 +312,44 @@ func TestDeleteHostsCommandFails(t *testing.T) {
...
+
+		err := NewCVDRemoteCommand(opts).Execute()
+
+		if _, ok := err.(MissingEnvVarErr); !ok {
+			t.Errorf("expected %+v, got %+v", MissingEnvVarErr(envVar), err)
+		}
+	}
+}
+
+func TestCreateCVDLocalImageMissingRequiredImagesFile(t *testing.T) {
```
jemoreira, line 338: This test depends on the fact that the code only checks the presence of the env variables before attempting to read the required file list, which is an implementation detail subject to change and not part of the function's contract. For example, if you were to check the existence of `cvd-host-packages.tar.gz` first and then try to read the list of images the test will begin failing.
Instead, you could test the `listRequiredImageFilenames` function directly by checking that a few files you added to a temporary "required files" file and `cvd-host-packages.tar.gz` are returned.

pkg/cli/cvd.go
```diff
@@ -178,3 +231,28 @@ func (o *CVDOutput) String() string {
 		fmt.Sprintf("%s/hosts/%s/cvds/%s/logs/", o.BaseURL, o.Host, o.CVD.Name)
 	return res
 }
+
+const cvdHostPackageName = "cvd-host_package.tar.gz"
+
+func listRequiredImageFilenames(listFilename, androidProductOutDir, androidHostOutDir string) ([]string, error) {
```
jemoreira, line 237: nit: I think it would make more sense to pass `androidBuildTop` to this function instead of `listFilename` and construct the latter in this function.

pkg/cli/cli.go
```diff
@@ -35,6 +35,8 @@ type CommandOptions struct {
 	IOStreams
 	Args   []string
 	Config Config
+	// Retrieves the value of the environment variable named by the key.
+	LookupEnv func(key string) (string, bool)
```
jemoreira, line 39: I get the feeling this variable only exists to allow testing. If that's the case you could take advantage of the fact that Go tests are never run in parallel unless you mark them so, which means it would be safe to set the environment variable before the test and resetting it back to its original state after.
Environment variables are  only available in the process that sets them and in any forked processes, which is why you need the guarantee that the test won't run concurrently with anything else and that no forks will be made.
This is more a suggestion than a request, if you take it make sure to add a comment in the test explaining the assumption about non-concurrency and warning against enabling it for that particular test.

pkg/cli/cvd.go
```diff
@@ -86,17 +91,65 @@ func newCVDCommand(cfgFlags *configFlags) *cobra.Command {
...
+	return fmt.Sprintf("Required images file not found at %q", string(s))
+}
+
+const (
+	AndroidBuildTopVarName   = "ANDROID_BUILD_TOP"
+	AndroidProductOutVarName = "ANDROID_PRODUCT_OUT"
+	AndroidHostOutVarName    = "ANDROID_HOST_OUT"
+)
+
+func createCVDFromLocalBuild(
```
jemoreira, line 125: nit: You could rewrite this function as
```
func createCVDFromLocalBuild(...) {
    // dirs is a struct with 3 string fields
    dirs, err := FileLocationsFromEnv()
    if err != nil {...}

    files, err := ListRequiredFilenames(dirs)
    if err != nil {...}

    if err := apiClient.UploadFiles(..., files); err != nil {...}
    return nil
}
```
That way you could test each of the 3 simpler functions on their own instead of the larger and more complicated `createCVDFromLocalBuild` one.

jemoreira (changes requested):

pkg/cli/cvd.go
ser-io: agree! done

pkg/cli/cvd.go
ser-io: good catch! this approach in favor of suggested os.ReadFile

pkg/cli/cvd.go
ser-io: agree, done

pkg/cli/cli_test.go
ser-io: agree, removed the test, added new exported method as suggested down below and added unit test to them

pkg/cli/cli_test.go
ser-io: agree, done

pkg/cli/cvd.go
ser-io: agree! done

pkg/cli/cli.go
ser-io: good point, used t.Setenv https://pkg.go.dev/testing#B.Setenv instead, documentation mentions that tests using this method cannot be executed in parallel.

pkg/cli/cvd.go
ser-io: definitely way more readable, thanks. done!

pkg/cli/cvd.go
ser-io: done

jemoreira (approved):

PR #90, opened 2023-01-11
[cvdremote] Allows to pass different `client.Service` implementations.

ser-io:
- Useful for unit testing.

### DIFF
--- a/cmd/cvdremote/main.go
+++ b/cmd/cvdremote/main.go
@@ -19,6 +19,7 @@ import (
 	"os"
 
 	"github.com/google/cloud-android-orchestration/pkg/cli"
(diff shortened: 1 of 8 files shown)


jemoreira (commented):
I can't figure it out from the github UI, but it looks like there is an empty commit? There should be only one in that case.

ser-io:
yes, I pushed an empty commit to trigger the presubmit actions which were not trigger with the initial commit, I'll Squash and merge onto the main repository, so the empty one won't be seen.

jemoreira:
Next time try `git commit --amend && git push -f` first instead of a new empty commit. The amend should change the commit message time and therefore the SHA which github should treat as a new push and re-trigger the checks.

jemoreira (approved):

PR #92, opened 2023-01-14
[cvdremote] Upload files.

ser-io:
- Uploading files will be done in chunks in order to be flexible with request limits set by app hosts providers. For example App Engine has a limit of 32MB https://cloud.google.com/appengine/docs/standard/quotas

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -412,22 +412,8 @@ func (u *filesUploader) Upload() error {
 	resultsChan := u.startWorkers(ctx, jobsChan)
 	go func() {
 		defer close(jobsChan)
-		for _, info := range infos {
-			for i := 0; i < info.TotalChunks; i++ {
-				select {
-				case <-ctx.Done():
-					return
-				default:
-					j := uploadChunkJob{
-						Filename:       info.Name,
-						ChunkNumber:    i + 1,
-						TotalChunks:    info.TotalChunks,
-						ChunkSizeBytes: u.ChunkSizeBytes,
-					}
-					jobsChan <- j
-				}
-			}
-		}
+		u.sendJobs(ctx, jobsChan, infos)
+
 	}()
 	// Only first error will be returned.
 	var returnErr error
@@ -461,6 +447,25 @@ func (u *filesUploader) getFilesInfos() ([]fileInfo, error) {
 	return infos, nil
 }
 
+func (u *filesUploader) sendJobs(ctx context.Context, jobsChan chan<- uploadChunkJob, infos []fileInfo) {
+	for _, info := range infos {
+		for i := 0; i < info.TotalChunks; i++ {
+			job := uploadChunkJob{
+				Filename:       info.Name,
+				ChunkNumber:    i + 1,
+				TotalChunks:    info.TotalChunks,
+				ChunkSizeBytes: u.ChunkSizeBytes,
+			}
+			select {
+			case <-ctx.Done():
+				return
+			case jobsChan <- job:
+				continue
+			}
+		}
+	}
+}
+
 func (u *filesUploader) startWorkers(ctx context.Context, jobsChan <-chan uploadChunkJob) <-chan error {
 	agg := make(chan error)
 	wg := sync.WaitGroup{}
@@ -472,12 +477,13 @@ func (u *filesUploader) startWorkers(ctx context.Context, jobsChan <-chan upload
 			DumpOut:     u.DumpOut,
 			JobsChan:    jobsChan,
 		}
-		go func(ch <-chan error) {
+		go func() {
+			defer wg.Done()
+			ch := w.Start()
 			for err := range ch {
 				agg <- err
 			}
-			wg.Done()
-		}(w.Start())
+		}()
 	}
 	go func() {
 		wg.Wait()
@@ -500,8 +506,6 @@ type uploadChunkWorker struct {
 	EndpointURL string
 	DumpOut     io.Writer
 	JobsChan    <-chan uploadChunkJob
-
-	bodyBuff bytes.Buffer
 }
 
 // Returns a channel that will return the result for each of the handled `uploadChunkJob` instances.
@@ -517,6 +521,46 @@ func (w *uploadChunkWorker) Start() <-chan error {
 }
 
 func (w *uploadChunkWorker) upload(job uploadChunkJob) error {
+	// return errors.New("abc")
+	pipeReader, pipeWriter := io.Pipe()
+	writer := multipart.NewWriter(pipeWriter)
+	go func() {
+		defer pipeWriter.Close()
+		defer writer.Close()
+		if err := writeMultipartRequest(writer, job); err != nil {
+			fmt.Fprintf(w.DumpOut, "Error writing multipart request %v", err)
+		}
+	}()
+	// client trace to log whether the request's underlying tcp connection was re-used
+	clientTrace := &httptrace.ClientTrace{
+		GotConn: func(info httptrace.GotConnInfo) {
+			if !info.Reused {
+				const msg = "tcp connection was not reused uploading file chunk: %q," +
+					"chunk number: %d, chunk total: %d\n"
+				fmt.Fprintf(w.DumpOut, msg,
+					filepath.Base(job.Filename), job.ChunkNumber, job.TotalChunks)
+			}
+		},
+	}
+	traceCtx := httptrace.WithClientTrace(context.Background(), clientTrace)
+	req, err := http.NewRequestWithContext(traceCtx, http.MethodPut, w.EndpointURL, pipeReader)
+	if err != nil {
+		return err
+	}
+	req.Header.Set("Content-Type", writer.FormDataContentType())
+	res, err := w.Client.Do(req)
+	if err != nil {
+		return err
+	}
+	if res.StatusCode != 200 {
+		const msg = "Failed uploading file chunk with status code %q. " +
+			"File %q, chunk number: %d, chunk total: %d."
+		return fmt.Errorf(msg, res.Status, filepath.Base(job.Filename), job.ChunkNumber, job.TotalChunks)
+	}
+	return nil
+}
+
+func writeMultipartRequest(writer *multipart.Writer, job uploadChunkJob) error {
 	file, err := os.Open(job.Filename)
 	if err != nil {
 		return err
@@ -525,9 +569,12 @@ func (w *uploadChunkWorker) upload(job uploadChunkJob) error {
 	if _, err := file.Seek(int64(job.ChunkNumber-1)*job.ChunkSizeBytes, 0); err != nil {
 		return err
 	}
-	writer := multipart.NewWriter(&w.bodyBuff)
-	addFormField(writer, "chunk_number", strconv.Itoa(job.ChunkNumber))
-	addFormField(writer, "chunk_total", strconv.Itoa(job.TotalChunks))
+	if err := addFormField(writer, "chunk_number", strconv.Itoa(job.ChunkNumber)); err != nil {
+		return err
+	}
+	if err := addFormField(writer, "chunk_total", strconv.Itoa(job.TotalChunks)); err != nil {
+		return err
+	}
 	fw, err := writer.CreateFormFile("file", filepath.Base(job.Filename))
 	if err != nil {
 		return err
@@ -541,30 +588,17 @@ func (w *uploadChunkWorker) upload(job uploadChunkJob) error {
 			return err
 		}
 	}
-	writer.Close()
-	// client trace to log whether the request's underlying tcp connection was re-used
-	clientTrace := &httptrace.ClientTrace{
-		GotConn: func(info httptrace.GotConnInfo) {
-			if !info.Reused {
-				const msg = "tcp connection was not reused uploading file chunk: %q," +
-					"chunk number: %d, chunk total: %d\n"
-				fmt.Fprintf(w.DumpOut, msg, job.Filename, job.ChunkNumber, job.TotalChunks)
-			}
-		},
-	}
-	traceCtx := httptrace.WithClientTrace(context.Background(), clientTrace)
-	req, err := http.NewRequestWithContext(
-		traceCtx, http.MethodPut, w.EndpointURL, bytes.NewReader(w.bodyBuff.Bytes()))
+	return nil
+}
+
+func addFormField(writer *multipart.Writer, field, value string) error {
+	fw, err := writer.CreateFormField(field)
 	if err != nil {
 		return err
 	}
-	req.Header.Set("Content-Type", writer.FormDataContentType())
-	res, err := w.Client.Do(req)
+	_, err = io.Copy(fw, strings.NewReader(value))
 	if err != nil {
 		return err
-	}
-	if res.StatusCode != 200 {
-		return errors.New(res.Status)
 	}
 	return nil
 }
@@ -584,17 +618,5 @@ func dumpResponse(r *http.Response, w io.Writer) error {
 		return err
 	}
 	fmt.Fprintf(w, "%s\n", dump)
-	return nil
-}
-
-func addFormField(writer *multipart.Writer, field, value string) error {
-	fw, err := writer.CreateFormField(field)
-	if err != nil {
-		return err
-	}
-	_, err = io.Copy(fw, strings.NewReader(value))
-	if err != nil {
-		return err
-	}
 	return nil
 }
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -29,6 +29,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"sync"
 	"time"
 
 	apiv1 "github.com/google/cloud-android-orchestration/api/v1"
@@ -88,7 +89,7 @@ type Service interface {
 
 	ListCVDs(host string) ([]*hoapi.CVD, error)
 
-	CreateUploadDir(host string) (string, error)
+	CreateUpload(host string) (string, error)
 
 	UploadFiles(host, uploadDir string, filenames []string) error
 }
@@ -303,7 +304,7 @@ func (c *serviceImpl) ListCVDs(host string) ([]*hoapi.CVD, error) {
 	return res.CVDs, nil
 }
 
-func (c *serviceImpl) CreateUploadDir(host string) (string, error) {
+func (c *serviceImpl) CreateUpload(host string) (string, error) {
 	uploadDir := &hoapi.UploadDirectory{}
 	if err := c.doRequest("POST", "/hosts/"+host+"/userartifacts", nil, uploadDir); err != nil {
 		return "", err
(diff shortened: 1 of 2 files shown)


pkg/client/client.go
```diff
@@ -352,6 +385,146 @@ func (c *serviceImpl) doRequest(method, path string, reqpl, respl any) error {
...
+	file, err := os.Open(job.Filename)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+	if _, err := file.Seek(int64(job.ChunkNumber-1)*job.ChunkSizeBytes, 0); err != nil {
+		return err
+	}
+	mpWriter := multipart.NewWriter(&w.bodyBuff)
+	name := fmt.Sprintf("%s_%d_%d.chunked", filepath.Base(job.Filename), job.ChunkNumber, job.TotalChunks)
```
jemoreira, line 487: This looks like the information about chunk number, size and count is included in the file name (and I guess the server will save them all with that name until all chunks are available and then concatenate them all). That adds unnecessary coupling between the server and the client while sending important data in a non-standard format.
Given that it's multipart form data, you can send whatever name-value pairs you want as form fields (not just files), for which Golang probably has nice ways of parsing into structs. So I'd prefer to send the chunk info as extra fields and the file name unmodified. The server can still choose to save the files with that naming scheme and concatenate at the end (although you can use a similar `open`, `seek`, `write` and `close` approach to write directly to the destination file, which will probably enable performance improvements of sparse files later)

pkg/client/client.go
```diff
@@ -352,6 +385,146 @@ func (c *serviceImpl) doRequest(method, path string, reqpl, respl any) error {
...
+	var totalChunks int
+	var infos []fileInfo
+	for _, name := range u.Filenames {
+		stat, err := os.Stat(name)
+		if err != nil {
+			return nil
+		}
+		info := fileInfo{
+			Name:        name,
+			TotalChunks: int(math.Ceil(float64(stat.Size()) / float64(u.ChunkSizeBytes))),
```
jemoreira, line 413: Use this formula with integer operations instead: `(file_size + chunk_size - 1) / chunk_size` 
Floating point operations are slow and approximate so it can introduce errors (though very unlikely in this case). The downside of the integer only operations is that the addition could potentially overflow, but that's not an issue for our use case with `int64`.

pkg/client/client.go
```diff
@@ -352,6 +385,146 @@ func (c *serviceImpl) doRequest(method, path string, reqpl, respl any) error {
...
+	if job.ChunkNumber < job.TotalChunks {
+		if _, err = io.CopyN(part, file, job.ChunkSizeBytes); err != nil {
+			return err
+		}
+	} else {
+		if _, err = io.Copy(part, file); err != nil {
+			return err
+		}
+	}
+	mpWriter.Close()
```
jemoreira, line 501: shouldn't this be on a `defer` above? Any errors writing would cause an early return and this close to be skipped.

pkg/client/client.go
```diff
@@ -352,6 +385,146 @@ func (c *serviceImpl) doRequest(method, path string, reqpl, respl any) error {
...
+	// client trace to log whether the request's underlying tcp connection was re-used
+	clientTrace := &httptrace.ClientTrace{
```
jemoreira, line 503: is this a leftover from debugging? If that dump line is intentional it should include more information, like chunk number out chunk count and file name. That way we'll get a better idea of how many connections were not reused.

pkg/client/client.go
```diff
@@ -352,6 +385,146 @@ func (c *serviceImpl) doRequest(method, path string, reqpl, respl any) error {
...
+			}
+		},
+	}
+	traceCtx := httptrace.WithClientTrace(context.Background(), clientTrace)
+	req, err := http.NewRequestWithContext(
+		traceCtx, http.MethodPut, w.EndpointURL, bytes.NewReader(w.bodyBuff.Bytes()))
+	if err != nil {
+		return err
+	}
+	req.Header.Set("Transfer-Encoding", "chunked")
```
jemoreira, line 516: What you're doing doesn't really match the [chunked description on wikipedia](https://en.wikipedia.org/wiki/Chunked_transfer_encoding#:~:text=Chunked%20transfer%20encoding%20is%20a,received%20independently%20of%20one%20another.). Is it really necessary here?

jemoreira (changes requested):

pkg/client/client.go
```diff
@@ -58,12 +64,13 @@ func (e *ApiCallError) Is(target error) bool {
...
-	ErrOut        io.Writer
-	RetryAttempts int
-	RetryDelay    time.Duration
+	BaseURL        string
+	ProxyURL       string
+	DumpOut        io.Writer
+	ErrOut         io.Writer
+	RetryAttempts  int
+	RetryDelay     time.Duration
+	ChunkSizeBytes int64
```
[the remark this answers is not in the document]
ser-io: Good point, however the relevant method in this PR are implemented in the host orchestrator, and the cloud orchestrator acts as a proxy which is the one that have the quota limits, we can add a new API method to the cloud orchestration service to get the quota limits and use it here later on.

pkg/client/client.go
```diff
@@ -292,6 +303,28 @@ func (c *serviceImpl) ListCVDs(host string) ([]*hoapi.CVD, error) {
...
+	uploadDir := &hoapi.UploadDirectory{}
+	if err := c.doRequest("POST", "/hosts/"+host+"/userartifacts", nil, uploadDir); err != nil {
+		return "", err
+	}
+	return uploadDir.Name, nil
+}
+
+func (c *serviceImpl) UploadFiles(host, uploadDir string, filenames []string) error {
+	if c.ChunkSizeBytes == 0 {
+		panic("ChunkSizeBytes value cannot be zero")
```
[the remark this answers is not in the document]
ser-io: I prefer to take a YAGNI stand here, it definitely could be implemented as you mentioned if needed.

pkg/client/client.go
```diff
@@ -352,6 +385,146 @@ func (c *serviceImpl) doRequest(method, path string, reqpl, respl any) error {
...
+	for i := 0; i < openConnections; i++ {
+		w := &uploadChunkWorker{
+			Client:      u.Client,
+			EndpointURL: u.EndpointURL,
+			DumpOut:     u.DumpOut,
+			JobsChan:    jobsChan,
+			ResultsChan: resultsChan,
+		}
+		go w.Start()
+	}
```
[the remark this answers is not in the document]
ser-io: don't understand the comment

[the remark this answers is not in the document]
jemoreira: Create a function named `startWorkers` and extract this section to it.

I thought I had left another comment about simplifying this function because it's too long but can't find it now so...: this function is too long and can be refactored into some calls to smaller functions (nit)

pkg/client/client.go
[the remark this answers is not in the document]
jemoreira: I see. I guess it's ok to get them from the configuration instead.

pkg/client/client.go
```diff
@@ -80,6 +87,10 @@ type Service interface {
 	CreateCVD(host string, req *hoapi.CreateCVDRequest) (*hoapi.CVD, error)
 
 	ListCVDs(host string) ([]*hoapi.CVD, error)
+
+	CreateUploadDir(host string) (string, error)
```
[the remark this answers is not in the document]
ser-io: Done, used `CreateUpload`

pkg/client/client.go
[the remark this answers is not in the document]
ser-io: yes, that's another option worth evaluating, I'll keep as a constant for now.

pkg/client/client.go
```diff
@@ -352,6 +385,146 @@ func (c *serviceImpl) doRequest(method, path string, reqpl, respl any) error {
...
+		if err != nil {
+			return nil
+		}
+		info := fileInfo{
+			Name:        name,
+			TotalChunks: int(math.Ceil(float64(stat.Size()) / float64(u.ChunkSizeBytes))),
+		}
+		infos = append(infos, info)
+		totalChunks += info.TotalChunks
+	}
```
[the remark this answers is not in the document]
ser-io: Done

pkg/client/client.go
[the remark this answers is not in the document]
ser-io: Done

pkg/client/client.go
```diff
@@ -352,6 +385,146 @@ func (c *serviceImpl) doRequest(method, path string, reqpl, respl any) error {
...
+	jobsChan := make(chan *uploadChunkJob, openConnections)
```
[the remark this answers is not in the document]
ser-io: great point, using values now.

pkg/client/client.go
```diff
@@ -352,6 +385,146 @@ func (c *serviceImpl) doRequest(method, path string, reqpl, respl any) error {
...
+	resultsChan := make(chan error, 1)
+	defer close(jobsChan)
```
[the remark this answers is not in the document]
ser-io: good point, done.

pkg/client/client.go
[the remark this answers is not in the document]
ser-io: true, the Host Orchestrator API does not have this method yet, but it will be added after finishing the upload work.

pkg/client/client.go
```diff
@@ -352,6 +385,146 @@ func (c *serviceImpl) doRequest(method, path string, reqpl, respl any) error {
 	return nil
 }
 
+const openConnections = 10
+
+type filesUploader struct {
```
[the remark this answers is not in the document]
ser-io: acknowledged.

pkg/client/client.go
```diff
@@ -352,6 +385,146 @@ func (c *serviceImpl) doRequest(method, path string, reqpl, respl any) error {
...
+			Name:        name,
+			TotalChunks: int(math.Ceil(float64(stat.Size()) / float64(u.ChunkSizeBytes))),
+		}
+		infos = append(infos, info)
+		totalChunks += info.TotalChunks
+	}
+	jobsChan := make(chan *uploadChunkJob, openConnections)
+	resultsChan := make(chan error, 1)
+	defer close(jobsChan)
+	defer close(resultsChan)
```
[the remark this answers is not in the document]
ser-io: great catch! discussed offline and the result was having a cancellation propagation solution where in case of error, the jobs channels get closed, then the workers's results channels and lastly the aggregated channel, following the rule of thumb mentioned above `channels should be closed from the side that writes to them`

pkg/client/client.go
```diff
@@ -352,6 +385,146 @@ func (c *serviceImpl) doRequest(method, path string, reqpl, respl any) error {
...
+		if err := <-resultsChan; err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+type uploadChunkJob struct {
+	Filename string
+	// Chunk numbers are 1-indexed.
```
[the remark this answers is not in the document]
ser-io: Elaborated a little bit more here as suggested putting emphasis on `number` rather than `index`.

pkg/client/client.go
ser-io: Good point! done

pkg/client/client.go
ser-io: Good point! done

pkg/client/client.go
ser-io: `close` here signals that http request performed at the end is completed, that there's nothing else to write. so deferring this would keep that http request open, in this case is fine if this writer is not closed in case of early return as well.

pkg/client/client.go
ser-io: added more info as suggested.

pkg/client/client.go
ser-io: not really, removed.

pkg/client/client.go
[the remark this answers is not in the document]
jemoreira: Please add a TODO then

pkg/client/client.go
```diff
@@ -352,6 +387,188 @@ func (c *serviceImpl) doRequest(method, path string, reqpl, respl any) error {
...
+func (w *uploadChunkWorker) upload(job uploadChunkJob) error {
+	file, err := os.Open(job.Filename)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+	if _, err := file.Seek(int64(job.ChunkNumber-1)*job.ChunkSizeBytes, 0); err != nil {
+		return err
+	}
+	writer := multipart.NewWriter(&w.bodyBuff)
```
jemoreira, line 528: This code reads the entire chunk into memory before sending it to the server. For smaller chunks this isn't a problem, but nothing guarantees the chunks will always be small, especially with big files such as `super.img`.
If you replace the in-memory buffer `w.bodyBuff` with a [pipe](https://pkg.go.dev/io#Pipe) that problem goes away at the cost of having to write the request body in an additional routine, but luckily you're already using a context to cancel the request if there are errors on that routine. Double check me on this, but if you cancel the request through the context `w.Client.Do(req)` will return an error which you can return so there aren't many changes required to make this work.

pkg/client/client.go
jemoreira: I see.
It's a shame they added a "Close" method to a "Writer" created from another "Writer" that doesn't "Close" the wrapped "Writer" (i.e that method should have a different name).

pkg/client/client.go
```diff
@@ -352,6 +387,188 @@ func (c *serviceImpl) doRequest(method, path string, reqpl, respl any) error {
...
+			EndpointURL: u.EndpointURL,
+			DumpOut:     u.DumpOut,
+			JobsChan:    jobsChan,
+		}
+		go func(ch <-chan error) {
+			for err := range ch {
+				agg <- err
+			}
+			wg.Done()
+		}(w.Start())
```
jemoreira, line 480: nit: I struggled to understand this block, consider rewriting as
```
go func() {
	var ch <-chan error = w.Start() // or just ch := w.Start()
	for err := range ch {
		agg <- err
	}
	wg.Done()
}()
```

jemoreira (changes requested):

pkg/client/client.go
ser-io: great point, used pipe instead as suggested. done

pkg/client/client.go
```diff
@@ -352,6 +387,188 @@ func (c *serviceImpl) doRequest(method, path string, reqpl, respl any) error {
 	return nil
 }
 
+const openConnections = 10
```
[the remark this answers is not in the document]
ser-io: true, I'd like to have it in a follow up PR though.

pkg/client/client.go
```diff
@@ -352,6 +387,188 @@ func (c *serviceImpl) doRequest(method, path string, reqpl, respl any) error {
...
+
+func (u *filesUploader) Upload() error {
+	infos, err := u.getFilesInfos()
+	if err != nil {
+		return err
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	jobsChan := make(chan uploadChunkJob)
+	resultsChan := u.startWorkers(ctx, jobsChan)
+	go func() {
```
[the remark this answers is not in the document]
ser-io: done

pkg/client/client.go
ser-io: good point, done

pkg/client/client.go
```diff
@@ -352,6 +387,188 @@ func (c *serviceImpl) doRequest(method, path string, reqpl, respl any) error {
...
+				case <-ctx.Done():
+					return
+				default:
+					j := uploadChunkJob{
+						Filename:       info.Name,
+						ChunkNumber:    i + 1,
+						TotalChunks:    info.TotalChunks,
+						ChunkSizeBytes: u.ChunkSizeBytes,
+					}
+					jobsChan <- j
```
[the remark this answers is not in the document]
ser-io: great catch, removed the blocking write from the `default` case

pkg/client/client.go
```diff
@@ -461,6 +447,30 @@ func (u *filesUploader) getFilesInfos() ([]fileInfo, error) {
...
+			for !msgSent {
+				select {
+				case <-ctx.Done():
+					return
+				case jobsChan <- job:
+					msgSent = true
+				default:
+					continue
+				}
+			}
```
jemoreira: The `default` in that `select` clause will cause it to do a "busy wait" for one of the channels to become ready, instead of just sleeping until it happens which would safe a lot of cpu cycles. I'd drop the loop and the boolean and just do
```
select {
case <-ctx.Done():
  return
case jobsChan <- job:
  // nothing
}
```

pkg/client/client.go
```diff
@@ -517,6 +521,46 @@ func (w *uploadChunkWorker) Start() <-chan error {
...
 
 func (w *uploadChunkWorker) upload(job uploadChunkJob) error {
+	// return errors.New("abc")
+	pipeReader, pipeWriter := io.Pipe()
+	writer := multipart.NewWriter(pipeWriter)
+	go func() {
+		defer pipeWriter.Close()
+		defer writer.Close()
+		if err := writeMultipartRequest(writer, job); err != nil {
+			fmt.Fprintf(w.DumpOut, "Error writing multipart request %v", err)
```
jemoreira, line 531: call the context's cancel on error copying the file.

That should cause the request to be cancelled instead of sent with incomplete/invalid data and wait for the server's error response.

pkg/client/client.go
```diff
@@ -517,6 +521,46 @@ func (w *uploadChunkWorker) Start() <-chan error {
...
+		GotConn: func(info httptrace.GotConnInfo) {
+			if !info.Reused {
+				const msg = "tcp connection was not reused uploading file chunk: %q," +
+					"chunk number: %d, chunk total: %d\n"
+				fmt.Fprintf(w.DumpOut, msg,
+					filepath.Base(job.Filename), job.ChunkNumber, job.TotalChunks)
+			}
+		},
+	}
+	traceCtx := httptrace.WithClientTrace(context.Background(), clientTrace)
```
jemoreira, line 545: Use the same context used for the workers instead of `context.Background()`: Cancelling the workers should also cancel the request if it's already started.

jemoreira (changes requested):

pkg/client/client.go
jemoreira: Done

pkg/client/client.go
ser-io: good idea, used a new context for each request though, this way different requests are not tied to the same context.

pkg/client/client.go
ser-io: used the per request context.

pkg/client/client.go
jemoreira: it's ok to use a per-request context, but the parent of those contexts needs to be the worker's context, not `context.Background()`.

jemoreira (changes requested):

pkg/client/client.go
ser-io: agree, done

jemoreira (approved):

PR #93, opened 2023-01-31
[cli] Create cvd with local image.

ser-io:


pkg/cli/cli.go
```diff
@@ -144,7 +144,7 @@ func buildServiceBuilder(builder client.ServiceBuilder) serviceBuilder {
 			ErrOut:         c.ErrOrStderr(),
 			RetryAttempts:  3,
 			RetryDelay:     5 * time.Second,
-			ChunkSizeBytes: 10 * 1024 * 1024,
+			ChunkSizeBytes: 16 * 1024 * 1024,
```
jemoreira, line 147: Please extract this change in constant value, along with the one in client.go to a separate commit and use the "rebase and merge" strategy instead of "squash and merge"

jemoreira (approved):

pkg/cli/cli.go
ser-io: done, in client.go we already use a constant variable.

PR #94, opened 2023-02-01
[CLI] Control sockets for ADB tunnels

jemoreira:


pkg/cli/adbtunnel.go
```diff
@@ -223,48 +236,256 @@ func (f *ADBForwarder) Send(data []byte) error {
...
+		ServiceURL: f.serviceURL,
+		Zone:       f.zone,
+		Host:       f.host,
+		Device:     f.device,
+		Port:       f.port,
+		Status:     status,
+	}
+	ret, err := json.Marshal(&msg)
+	if err != nil {
+		panic("Couldn't marshal status map")
```
ser-io, line 269: nit: add the error text to this message.

pkg/cli/adbtunnel.go
```diff
@@ -223,48 +236,256 @@ func (f *ADBForwarder) Send(data []byte) error {
...
-			// listener was closed, ignore it.
+		conn, err := f.listener.Accept()
+		if err != nil {
+			if !errors.Is(err, net.ErrClosed) {
+				f.logger.Printf("Error accepting connection on port %d: %v", f.port, err)
+			}
 			return
 		}
+		f.statusMtx.Lock()
 		f.conn = conn
```
ser-io, line 298: nit: extract this into a new smaller method: `setConnection`

pkg/cli/adbtunnel.go
```diff
@@ -223,48 +236,256 @@ func (f *ADBForwarder) Send(data []byte) error {
...
+	// This will cause the control loop to finish.
+	tc.control.Close()
+}
+
+func (tc *TunnelController) StartControlLoop(wg *sync.WaitGroup) {
+	wg.Add(1)
+	go tc.controlSocketLoop(wg)
+}
+
+func (tc *TunnelController) GetForwarderPort() int {
```
ser-io, line 409: nit: `ForwarderPort()` following naming convention https://google.github.io/styleguide/go/decisions#getters

pkg/cli/config.go
```diff
@@ -29,15 +29,15 @@ type HostConfig struct {
 	GCP GCPHostConfig
 }
 
-type Config struct {
+type FileConfig struct {
```
ser-io, line 32: Adding the `File` prefix seems  an implementation detail, where the config is pulled from. This struct represents the overall configuration of cvdremote before any command gets executed no matter how it's parameters are set.

pkg/cli/adbtunnel.go
```diff
@@ -94,11 +94,11 @@ func runOpenADBTunnelCommand(config Config, flags *openADBTunnelFlags, c *comman
 		adbPath = path
 	}
 
-	if config.ADBControlDir == "" {
-		config.ADBControlDir = DefaultADBControlDir
+	if opts.Config.ADBControlDir == "" {
```
ser-io, line 97: I was thinking that this logic is better put in config.go, in a step that fills out empty values with these defaults after they are pulled from file.

pkg/cli/config.go
jemoreira: Agreed, but just `Config` sounds like the union of the file config and the flags.
Accepting suggestions for a different name.

pkg/cli/adbtunnel.go
```diff
@@ -91,112 +99,117 @@ func runOpenADBTunnelCommand(flags *openADBTunnelFlags, c *command, args []strin
...
-			adbPath: adbPath,
+		logger := log.New(c.ErrOrStderr(), "", log.LstdFlags)
+		devProps := deviceProperties{
+			serviceURL: flags.ServiceURL,
+			zone:       flags.Zone,
+			host:       flags.host,
+			device:     device,
 		}
-		conn, err := service.ConnectWebRTC(flags.host, device, &observer)
+		controller, err := NewTunnelController(service, devProps, logger, &wg)
```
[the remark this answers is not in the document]
jemoreira: good point.

Called the loop directly from inside a routine created in the same function as where the wait group is created. The wait group is only used from that routine after the loop returns, so now there is no chance for mishandling.

pkg/cli/adbtunnel.go
jemoreira: Done
Great suggestion! I also added `compareAndSwapStatus(old, new)` to change states atomically.

pkg/cli/adbtunnel.go
jemoreira: Done

pkg/cli/adbtunnel.go
jemoreira: Done

pkg/cli/config.go
jemoreira: Done

pkg/cli/adbtunnel.go
jemoreira: Done

pkg/cli/adbtunnel.go
```diff
@@ -160,19 +165,36 @@ type ADBForwarder struct {
...
 	logger    *log.Logger
 }
 
 const (
-	ADBFwdInitializing = "initializing"
-	ADBFwdReady        = "ready"
-	ADBFwdConnected    = "connected"
-	ADBFwdStopped      = "stopped"
-	ADBFwdFailed       = "failed"
+	ADBFwdInitializing = 0 // The initial state after creation
```
ser-io, line 169: this seems a good use case to use https://go.dev/ref/spec#Iota

pkg/cli/adbtunnel.go
```diff
@@ -252,38 +277,53 @@ type StatusMsg struct {
...
-		panic("Couldn't marshal status map")
+		panic(fmt.Sprintf("Couldn't marshal status map: %v", err))
 	}
 	return ret
 }
 
-func (f *ADBForwarder) setStatus(status string) {
+// Changes f.status to newStatus if it had oldStatus. Returns whether the change was
+// made and the old status.
+func (f *ADBForwarder) compareAndSwapStatus(oldStatus, newStatus int) (bool, int) {
```
ser-io, line 295: nit: `old` and `new`?

ser-io, line 295: nit: use `current` instead of `old`, the method name could be: `swapStatusIfCurrent(current, new int)`

pkg/cli/adbtunnel.go
jemoreira: Done

pkg/cli/adbtunnel.go
jemoreira: compare_and_swap(old, new) is sort of a convention: https://en.wikipedia.org/wiki/Compare-and-swap

ser-io: I see, never mind then.

jemoreira: Done

ser-io (approved):

PR #96, opened 2023-02-02
Connect ADB through ADB server control socket instead of via command

jemoreira:
This avoids restarting the server unintentionally and eliminates the need to control the connection with a flag.

### DIFF
--- a/pkg/cli/adbtunnel.go
+++ b/pkg/cli/adbtunnel.go
@@ -32,33 +32,20 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const connectFlag = "connect"
-
 type ADBTunnelFlags struct {
 	*CommonSubcmdFlags
 	host string
 }
 
-type OpenADBTunnelFlags struct {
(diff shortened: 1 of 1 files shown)


pkg/cli/adbtunnel.go
```diff
@@ -509,10 +480,23 @@ func createControlSocket(dir string, port int) (*net.UnixListener, error) {
...
+func ADBConnect(adbSerial string) error {
+	const ADBServerPort = 5037
+	conn, err := net.Dial("tcp", fmt.Sprintf("127.0.0.1:%d", ADBServerPort))
+	if err != nil {
+		return fmt.Errorf("Unable to contact ADB server: %w", err)
+	}
+	defer conn.Close()
+	msg := fmt.Sprintf("host:connect:%s", adbSerial)
+	msg = fmt.Sprintf("%.4x%s", len(msg), msg)
+	fmt.Println(msg)
```
ser-io, line 492: nit: was this left from debuggin?

PR #98, opened 2023-02-02
[cli] - Make gcp machine type part of default initial config.

ser-io:


pkg/cli/config.go
```diff
@@ -40,6 +40,11 @@ type Config struct {
 func DefaultConfig() Config {
 	return Config{
 		ADBControlDir: "~/.cvdremote/adb",
+		Host: HostConfig{
+			GCP: GCPHostConfig{
+				MachineType: "n1-standard-4",
```
jemoreira, line 45: I'm not convinced this move is a good idea. Mostly because it's setting the default in stone (code). It seems perfectly reasonable to me to require a default to be set in the config file or require the user to specify it in the cmdline args, or even depend on a default at the service API level.

For comparison, the `ADBControlDir` is not configurable by cmdline args (so that `list` and `close` work correctly) and if a default is not provided the adbtunnel feature won't work. Additionally, the value provided for it is valid in (almost) all environments, as the concept of a home directory is practically universal. Even with those reasons in mind, I consider placing that default in the code to be an acceptable compromise, not the ideal solution.

ser-io: sounds good, I'm not using a config file in my development workflow yet, so I had to set the flag which was not the previous behavior,  I agree this is better handled in the config file, will revert this PR

PR #99, opened 2023-02-03
Implement adbtunnel list

jemoreira:


pkg/cli/adbtunnel.go
```diff
@@ -108,6 +117,40 @@ func openADBTunnel(flags *ADBTunnelFlags, c *command, args []string, opts *subCo
 	return merr
 }
 
+func listADBTunnels(flags *listADBTunnelFlags, c *command, args []string, opts *subCommandOpts) error {
+	controlDir := opts.InitialConfig.ADBControlDirExpanded()
+	entries, err := os.ReadDir(controlDir)
+	if err != nil {
+		return fmt.Errorf("Error reading control directory (%s): %w", controlDir, err)
```
ser-io, line 124: nit: %q will print a quotes string

pkg/cli/adbtunnel.go
```diff
@@ -449,6 +492,26 @@ func (tc *TunnelController) handleControlCommand(conn net.Conn) {
 	}
 }
 
+// The returned status is only valid if the error is nil
+func sendStatusCmd(conn net.Conn) (ADBForwarderStatus, error) {
+	var msg ADBForwarderStatus
+	// unixseqpacket means we don't have to write or read in a loop, since all
```
ser-io, line 503: is this comment more relevant close to `conn, err := net.Dial("unixpacket", path)` in line 133

pkg/cli/adbtunnel.go
```diff
@@ -37,6 +37,11 @@ type ADBTunnelFlags struct {
 	host string
 }
 
+type listADBTunnelFlags struct {
+	*ADBTunnelFlags
+	longFormat bool
```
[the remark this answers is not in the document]
jemoreira: Done

pkg/cli/adbtunnel.go
jemoreira: paths are not typically printed with quotes, changed the message to not need parenthesis either.

pkg/cli/adbtunnel.go
jemoreira: I'm trying to explain why I don't write in a loop here, as is customary with `int write(buffer)`, so the comment is more relevant here. Rewrote it to try to make it clearer.

ser-io (approved):

ser-io (approved):

PR #102, opened 2023-02-04
[cli] Refactors create cvd logic.

ser-io:


pkg/cli/cli_test.go
```diff
@@ -170,3 +171,14 @@ func newTestIOStreams() (IOStreams, *bytes.Buffer, *bytes.Buffer) {
...
+
+func cvdOutput(serviceURL, host, cvd string) string {
+	output := CVDOutput{
+		BaseURL: serviceURL + "/v1",
+		Host:    "foo",
+		CVD: &hoapi.CVD{
+			Name: "cvd-1",
+		},
+	}
+	return output.String() + "\n"
```
jemoreira, line 183: nit: if you had function that lists cvds and returns `[]CVDOutput` you wouldn't need to play with these string manipulations.

pkg/cli/cvd.go
```diff
@@ -87,29 +91,54 @@ func newCVDCommand(opts *subCommandOpts) *cobra.Command {
...
+	output := CVDOutput{
+		BaseURL: buildBaseURL(flags.CVDRemoteFlags),
+		Host:    flags.Host,
+		CVD:     cvd,
+	}
+	c.Printf("%s\n", output.String())
+	return nil
+}
+
+type createCVDAction struct {
```
jemoreira, line 114: nit: Instead of `createCVDAction` and `Execute`, you could have `CVDCreator` and `Create` respectively.

pkg/cli/cvd.go
```diff
@@ -87,29 +91,54 @@ func newCVDCommand(opts *subCommandOpts) *cobra.Command {
...
+	Service client.Service
+	Opts    CreateCVDOpts
+}
```
jemoreira, line 117: I like the idea of the extra class to break the operation into smaller pieces, but I think you could go even further by having:
* A `CVDCreator` interface with a `Create` method
* A `CVDCreator` builder (class, function, whatever makes the most sense) that parses the flags and instantiates the appropriate `CVDCreator` implementation
* Different implementations of the `CVDCreator` interface for local builds and android ci builds

jemoreira (changes requested):

pkg/cli/cli_test.go
ser-io: agree, there's room for refactoring here which I'd like to do in a different PR as it would be reused in other parts of the code.

pkg/cli/cvd.go
ser-io: `CVDCreator` seems appropriate, however the application layer will have many action so I was looking for a naming convention for the classes in the frontier of the application layer, i.e `listCVDsAction`, `deleteCVDAction`, `deleteHostAction` and so on.

pkg/cli/cvd.go
ser-io: agree, however as of right now the creation logic is quite simple, so having a new interface and two different creator classes seems overwhelming to me. I'll keep this in mind if at least of the creators logic starts getting bigger.

pkg/cli/cvd.go
jemoreira: Unless you are going to put a bunch of different actions together in a collection and therefore make them all implement an `Action` interface; there is no need to have them all share that name. Introducing polymorphism here doesn't seem to help, but it still has the disadvantage of reducing readability (example: `Execute` is not as readable as `CreateCVD`).

pkg/cli/cli_test.go
ser-io: went ahead and removed the string manipulations in the tests.

ser-io: done

pkg/cli/cvd.go
ser-io: acknowledged, done

jemoreira (approved):

PR #106, opened 2023-02-10
[cli] Respect the order of hosts when printing cvds.

ser-io:


pkg/cli/cvd.go
```diff
@@ -238,15 +242,62 @@ func listCVDs(c *cobra.Command, flags *ListCVDsFlags, opts *subCommandOpts) erro
 					fmt.Errorf("lists cvds for host %q failed: %w", name, err))
 				return
 			}
-			for _, cvd := range hostCVDs {
-				printCVDs(c.OutOrStdout(), buildBaseURL(flags.CVDRemoteFlags), name, []*hoapi.CVD{cvd})
-			}
-		}(host)
+			out := &bytes.Buffer{}
+			printCVDs(out, buildBaseURL(flags.CVDRemoteFlags), name, hostCVDs)
+			b, _ := io.ReadAll(out)
```
jemoreira, line 247: this seems more complicated than necessary: why not have a function that returns a string and have the caller print it however it wants?

pkg/cli/cvd.go
```diff
@@ -226,9 +229,10 @@ func listCVDs(c *cobra.Command, flags *ListCVDsFlags, opts *subCommandOpts) erro
 	var wg sync.WaitGroup
 	var mu sync.Mutex
 	var merr error
-	for _, host := range hosts {
+	sortedWriter := newSortedWriter(c.OutOrStdout())
```
jemoreira, line 232: two comments in one:
1- (nit) Don't like the overly complex use of mutexes and wait groups
2- Possible high latency from waiting for all calls to succeed before printing anything.

Both can be addressed by creating a channel per routine and having each routine write the result (message or error) to it. Then once all routines have been started the main routine reads from each channel in the right order and prints accordingly.

jemoreira (changes requested):

pkg/cli/cvd.go
ser-io: true, this was removed after addressing the other comment

pkg/cli/cvd.go
ser-io: great point, rephrased with channels.

pkg/cli/cvd.go
```diff
@@ -226,76 +228,31 @@ func listCVDs(c *cobra.Command, flags *ListCVDsFlags, opts *subCommandOpts) erro
...
-}
-
-// Writes payloads in order.
-type sortedWriter struct {
-	w           io.Writer
-	payloadsMap map[int]string
-	currIndex   int
-	mu          sync.Mutex
-}
+			ch <- result[[]*hoapi.CVD]{Result: cvds}
```
jemoreira, line 241: nit: `ch <- result[[]*hoapi.CVD]{cvds, err}` would avoid the `if`

pkg/cli/cvd.go
```diff
@@ -209,6 +206,11 @@ func (c *cvdCreator) createCVDFromAndroidCI() (*hoapi.CVD, error) {
 	return c.Service.CreateCVD(c.Opts.Host, &req)
 }
 
+type result[T any] struct {
```
jemoreira, line 209: nit: given that this is only useful for sending over channels and nothing else and that there is only one usage of this I wouldn't make it generic: it's only adding boilerplate for now. If still prefer having the generic version, add a type alias like `type CVDListResult result[[]*hoapi.CVD]`

jemoreira (approved):

pkg/cli/cvd.go
ser-io: agree, done

pkg/cli/cvd.go
ser-io: agree, done

PR #111, opened 2023-02-13
[cli] Fixes error.

ser-io:
- Passes the value of the newly created instance rather than the empty value from the command line.


jemoreira (approved):
nit: It's not immediately clear from the code what error is being fixed, it'd be better if the PR description and/or the commit message described it.

PR #114, opened 2023-02-15
Prevent concurrent connections to remote cvds

jemoreira:


pkg/cli/adbtunnel.go
```diff
@@ -233,24 +233,31 @@ func adbTunnelAgentLoop(flags *ADBTunnelFlags, c *command, args []string, opts *
...
+		Zone:       flags.Zone,
+		Host:       flags.host,
+		Device:     device,
 	}
-	logger, err := createLogger(opts.InitialConfig.ADBControlDirExpanded(), devProps)
-	if err != nil {
+
+	controlDir := opts.InitialConfig.ADBControlDirExpanded()
+	port, controller, err := findOrCreateTunnel(controlDir, devSpec, service)
+	if port < 0 {
```
ser-io, line 245: can we handle this using type assertion instead, i.e  if err.(FindOrCreateTunnelFatalError)?

jemoreira: separated non-fatal from fatal errors instead.

pkg/cli/adbtunnel.go
```diff
@@ -368,16 +367,18 @@ func gatherADBTunnels(controlDir string, zone, host string) ([]ADBForwarderStatu
 	return forwarders, merr
 }
 
-// Finds an existing tunnel or creates a new one.
-// Returns the port for the existing or created tunnel, the tunnel controller if
-// one was created and/or an error. A negative port number indicates failure, errors
-// may still be returned on success and could be presented to the user as warnings.
-func findOrCreateTunnel(controlDir string, devSpec DeviceSpec, service client.Service) (int, *TunnelController, error) {
+type findOrCreateRet struct {
```
ser-io, line 370: nit: properties here could be public.

pkg/cli/adbtunnel.go
```diff
@@ -368,16 +367,18 @@ func gatherADBTunnels(controlDir string, zone, host string) ([]ADBForwarderStatu
...
+	port        int
+	controller  *TunnelController
+	nonFatalErr error
```
ser-io, line 373: nit: could just `error` be fine here? the `nonFatal` is an implementation detail, users of this structure  are responsible of treating `t.error` as a non fatal error.

pkg/cli/adbtunnel.go
jemoreira: Done

pkg/cli/adbtunnel.go
jemoreira: Done

ser-io (approved):

PR #119, opened 2023-02-21
Do not assert on number of retries as it's an implementation detail.

ser-io:


pkg/client/client_test.go
```diff
@@ -223,12 +223,12 @@ func TestUploadFilesExponentialBackoffReachedElapsedTime(t *testing.T) {
...
 
 	err := srv.UploadFiles("foo", "bar", []string{waldoFile})
+
 	if err == nil {
 		t.Fatal("expected error")
 	}
-	expRetries := 4
-	if retries != expRetries {
-		t.Fatalf("expected %d, got %d", expRetries, retries)
+	if retries == 0 {
```
jemoreira, line 230: nit: attempts instead of retries then?

jemoreira (approved):

pkg/client/client_test.go
ser-io: sure, done

PR #124, opened 2023-02-23
Remove actions publishing the cvdremote binary.

ser-io:


.github/workflows/cvdremote_linux_continuous.yaml
```diff
@@ -8,7 +8,7 @@ on:
 jobs:
   build-linux:
     name: cvdremote-build-linux
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
```
jemoreira, line 11: The whole point of running in ubuntu 18.04 was to ensure **that** version remained usable, is there no other way around this?

ser-io: removed these actions, users can use `go tool` to install `cvdr` rather than downloading it.

jemoreira (approved):

PR #127, opened 2023-02-28
Overhaul client side ADB connections

jemoreira:
Connect to device on creation, with an option to [re]connect to existing devices and disconnect without destroying.

Refer to those connections as general connections instead of ADB specific. Will make it simpler to add more features later (fastboot).

List connections as part of `cvd list` and not by themselves (a single connection per CVD is allowed anyways).


pkg/cli/cli.go
```diff
@@ -34,11 +33,21 @@ type IOStreams struct {
 	ErrOut io.Writer
 }
 
+type CommandRunner interface {
+	// Start a command and doesn't wait for it to exit. Instead it reads its entire
+	// standard output and returns that or an error. The commands stdin and stderr
```
ser-io, line 39: isn't there a conflict between not waiting for it to exit and reading the entire standard output?

pkg/cli/cli.go
```diff
@@ -536,18 +540,23 @@ func runOpenADBTunnelCommand(flags *ADBTunnelFlags, c *command, args []string, o
...
 			return -1
 		}
+
 		return port
 	}
 
-	tunnelReporter := func(device string, port int) {
+	return openADBTunnels(args, agentLauncher, connReporter)
+}
+func runConnectCommand(flags *ADBTunnelFlags, c *command, args []string, opts *subCommandOpts) error {
```
ser-io: nit: leave an empty line above.

pkg/cli/cli.go
```diff
@@ -479,6 +479,22 @@ func runDeleteHostsCommand(c *cobra.Command, args []string, flags *CVDRemoteFlag
 	if err != nil {
 		return err
 	}
+	// Close connections first to avoid spurious error messages later.
+	for _, host := range args {
```
ser-io, line 463: nit: can this be moved into its own function?

pkg/cli/cvd.go
```diff
@@ -25,27 +25,28 @@ import (
...
-func (o *CVDOutput) String() string {
-	res := fmt.Sprintf("%s (%s)", o.CVD.Name, o.Host)
-	res += "\n  " + "Status: " + o.CVD.Status
-	adbState := ""
-	if o.ADBPort > 0 {
-		adbState = fmt.Sprintf("127.0.0.1:%d", o.ADBPort)
-	} else {
-		adbState = "disconnected"
+type CVD struct {
+	ServiceRootEndpoint string `json:"service_root_url"`
```
ser-io, line 29: nit: `service_root_endpoint` instead.

pkg/cli/cli.go
jemoreira: Done

pkg/cli/cli.go
jemoreira: Done

pkg/cli/cvd.go
jemoreira: Done

pkg/cli/cli.go
jemoreira: A background process (daemon) usually closes its standard file descriptors (https://netzmafia.ee.hm.edu/skripten/unix/linux-daemon-howto.html). It's common for the parent process launching a daemon to ensure that it started correctly and the easiest way to do that is to read from a pipe shared between both processes. The fact that said pipe is the daemon's stdout is a minor detail for us.
BTW this is the same thing `launch_cvd --daemon` does, except that it uses a different fd than stdout IIRC.

ser-io: I see, yeah I asked because cvd_server does not close the file descriptors so the execution gets blocked there https://github.com/google/android-cuttlefish/blob/main/frontend/src/host_orchestrator/orchestrator/instancemanager.go#L577-L582. I wasn't aware that it's expected for good "dameon" citizens to close its standard file descriptors.

ser-io (approved):

pkg/cli/cli.go
jemoreira: good daemons actually do `fork()` + `exec()` so you can actually just wait for them to exit and it's that daemon initial process that waits for some signal from the child to exit itself instead of waiting for it...
But that's impossible with Go because `fork()` only recreates the main thread and that breaks the runtime.

PR #129, opened 2023-03-01
Connection agent closes the command's IO channel instead of the global ones

jemoreira:
instead of the global channels from os.Std*.

This makes it more test-friendly

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -18,7 +18,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"os"
 	"time"
 
 	"github.com/google/cloud-android-orchestration/pkg/client"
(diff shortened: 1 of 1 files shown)


pkg/cli/cli.go
```diff
@@ -663,11 +662,17 @@ func runConnectionAgentCommand(flags *ConnectFlags, c *command, args []string, o
...
 	}
 
-	// Signal the caller that the agent is moving to the background.
-	// Ideally, this should close the command's streams, but those are not closeable.
-	os.Stdin.Close()
-	os.Stdout.Close()
-	os.Stderr.Close()
+	// Signal the caller that the agent is moving to the background by closing
+	// the command's standard IO channels.
+	if cin, closeable := c.InOrStdin().(io.Closer); closeable {
```
ser-io, line 667: nit: it's fine to use `ok` here to keep it consistent with the use of the `ok idiom` in the code base.

ser-io (changes requested):

ser-io (approved):

pkg/cli/cli.go
jemoreira: Done

PR #131, opened 2023-03-02
Add more functionality to cvdr connect and mark version 0.2.0 for unstable release

jemoreira:


pkg/cli/cli.go
```diff
@@ -395,35 +388,35 @@ func cvdCommands(opts *subCommandOpts) []*cobra.Command {
 }
 
 func connectionCommands(opts *subCommandOpts) []*cobra.Command {
-	connectFlags := &ConnectFlags{opts.RootFlags, ""}
+	connFlags := &ConnectFlags{opts.RootFlags, "", false /*skipConfirmation*/}
```
ser-io, line 391: nit: let's use `field: value` declaration to initialize here, I think it still could be in one line.

ser-io (changes requested):
let's move `ConnectDevice`, and `gatherDisconnectedDevices` to a different file keeping cli.go about commands layout.

jemoreira:
> let's move ConnectDevice, and gatherDisconnectedDevices to a different file keeping cli.go about commands layout.

Done

pkg/cli/cli.go
```diff
@@ -232,6 +227,35 @@ func PromptSelection[T any](c *command, choices []T, toStr func(T) string, selOp
 	return choices, nil
 }
 
+func PromptSelectionFromMap[K comparable, T any](c *command, choices map[K]T, toStr func(K, T) string, selOpt SelectionOption) (map[K]T, error) {
```
ser-io, line 230: let's add some comment about how this prompt looks like to understand when it can be used.

ser-io (approved):

pkg/cli/cli.go
jemoreira: done

PR #134, opened 2023-03-10
Create connection directories before using them

jemoreira:


pkg/cli/conn.go
```diff
@@ -53,6 +53,17 @@ func ControlSocketName(_ CVD, cs ConnStatus) string {
 	return fmt.Sprintf("%d.sock", cs.ADB.Port)
 }
 
+func EnsureConnDirsExist(controlDir string) error {
```
ser-io, line 56: nit: just `dir` here?

ser-io (approved):

pkg/cli/conn.go
jemoreira: there is more than one: the control sockets directory and the logs directory, the latter inside the former

PR #138, opened 2023-04-04
[orchestrator] Makes the web static files path configurable.

ser-io:


conf.toml
```diff
@@ -1,11 +1,13 @@
...
+
 [AccountManager]
 Type = "unix"
 
 [InstanceManager]
 Type = "unix"
 
 [InstanceManager.GCP]
-ProjectId = ""
+ProjectID = ""
```
jemoreira, line 10: This doesn't seem related to this change

pkg/app/signalingserver.go
```diff
@@ -32,11 +32,15 @@ import (
 // communicating with the host orchestrator in the appropriate instance and
 // forwarding all requests to it.
 type ForwardingSignalingServer struct {
+	interceptPath   string
```
jemoreira, line 35: nit: `interceptedFilesPath` is more clear

ser-io: used `connectorStaticFilesPath` as discussed offline.

conf.toml
ser-io: true, reverted.

PR #139, opened 2023-04-07
[orchestrator] Logs GCP Project Id if set.

ser-io:


cmd/cloud_orchestrator/main.go
```diff
@@ -33,6 +34,12 @@ func main() {
 	if err != nil {
 		log.Fatal("Failed to load configuration: ", err)
 	}
+	b, err := json.MarshalIndent(config, "", "    ")
+	if err != nil {
+		log.Fatal("Failed to marshal configuration:", err)
+	}
+	log.Println("Main Configuration:\n" + string(b))
```
jemoreira, line 41: This is dangerous if the configuration includes any secrets.

Even if we don't have any secrets in the configuration now, we could add them in the future or others may. It's better to just print the fields you're interested in.

jemoreira (changes requested):

cmd/cloud_orchestrator/main.go
ser-io: The config file is checked in in the code repository to be version controlled, so these should not include any secret. Secrets should be deployed in a different way.

ser-io: However, logging a full object is still a bad practice so I'll print specific fields as suggested.

ser-io: Done

jemoreira (approved):

PR #141, opened 2023-04-19
Allow https communication between cloud and host orchestrator

jemoreira:


pkg/app/unix/instancemanager.go
```diff
@@ -26,6 +26,9 @@ import (
 // device in the local host orchestrator.
 // This implementation is useful for both development and testing
 type InstanceManager struct{}
+func NewInstanceManager(cfg app.IMConfig) *InstanceManager {
```
ser-io, line 34: nit: leave an empty line above.

conf.toml
```diff
@@ -5,6 +5,8 @@ Type = "unix"
 
 [InstanceManager]
 Type = "unix"
+HostPort = 1080
```
ser-io: nit: for the future: the prefix Host here, from the context of the Cloud Orchestator Instance Manager is confusing, ideally it would HostOrchestratorServicePort and HostOrchestratorServiceProtocol, unfortunately these names are quite verbose, just something to think about it.

conf.toml
```diff
@@ -5,8 +5,9 @@ Type = "unix"
 
 [InstanceManager]
 Type = "unix"
-HostPort = 1080
-HostProtocol = "http"
+HostPort = 1443
```
ser-io, line 8: I think we need a new Setting class for these parameters, it could be [HostCommunication]. Having these parameters here is affecting the implementation bloating up the Instance Manager interface and I think the two concepts could be separated. We keep the IM responsibility for CRUD operations and a new component to deal with the communication challenges which is cloud provider agnostic.

pkg/app/gcp/instancemanager.go
```diff
@@ -204,6 +207,22 @@ func (m *InstanceManager) WaitOperation(zone string, user app.UserInfo, name str
 	return getter.Get()
 }
 
+func (m *InstanceManager) BuildHostReverseProxy(zone, host string) (*httputil.ReverseProxy, error) {
```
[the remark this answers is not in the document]
jemoreira: My reasoning was:
- The IM knows how to connect to the instances it manages
- The IM knows whether the certificates should be verified or not (returning a URL doesn't provide this information)
- The current setup provides a direct network connection between the Cloud Orchestrator and the host machines, but that is not required. The connection could go through an agent/proxy, for example, a URL won't work for that either. The current setup actually uses an agent of sorts, the serverless VPC connector, it just happens to be at the network layer.

Returning a client that can send HTTP requests (GET and POST for now, the other ones could be added if needed) to the HO seems like the best abstraction to me: It hides as many details as possible from the underlying architecture and only lets users of the interface do HTTP. Another example that a URL won't work for is if you deploy the entire system is a single server and decide to use UNIX sockets to communicate between CO and HO.

That being said, returning a reverse proxy does feel like a **hack** and I added it because there is no convenient way to create one from an HTTP client. I think it could be cleaner if we move the `build-reverse-proxy` method to the client interface, WDYT?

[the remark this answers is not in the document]
jemoreira: In case it wasn't evident: I intentionally removed the `GetURL` method from the interface.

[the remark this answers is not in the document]
ser-io: We can continue the discussion here: 
https://github.com/google/cloud-android-orchestration/pull/141/files/a4c8a69e00f533f94885859fb83551a9f5f29225#r1171699696.

From there I mention that we should only keep `GetHostAddr` in the IM interface get rid of the `GetHostURL`, `BuildHostReverseProxy` and `GetHostHTTPClient` from here and put the logic behind these methods in a different component cloud provider agnostic.

I don't argue against `GetHostURL` vs `GetHostHTTPClient`, now that we have a more complex setup, I think these two could be merged into just one, but in a different component.

pkg/app/unix/instancemanager.go
jemoreira: Done

conf.toml
ser-io: The way I see is is having the IM only responsible to provide the IP which is cloud provider specific, then we can have a new  common layer that is responsible for the http communication with the host orchestator, building the transport object, the http client and the reverse proxy. Putting it into the IM interface makes it bigger and I don't see a hard constrain for this.

pkg/app/gcp/instancemanager.go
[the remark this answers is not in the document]
jemoreira: That link doesn't work for me.

conf.toml
jemoreira: You're assuming that the Cloud Orchestrator only needs an IP address and port to communicate with the Host Orchestrator, I'm saying this won't always be the case.
If you want to have a class that specializes in CRUD and another one that specializes in comms, I'm ok with that, but those classes will be linked together at the implementation level, not the interface level. If you specify at the interface level the way you can reach the HO (by returning an IP and port) you'll be adding an unnecessary constraint.

pkg/app/controller.go
```diff
@@ -127,34 +130,28 @@ func (f *HostForwarder) Handler() HTTPHandler {
...
-	innerProxy *httputil.ReverseProxy
-}
-
-func newHostReverseProxy(hostName string, hostBaseURL *url.URL) *hostReverseProxy {
-	director := func(r *http.Request) {
-		r.URL.Scheme = hostBaseURL.Scheme
-		r.URL.Host = hostBaseURL.Host
-		split := strings.SplitN(r.URL.Path, "hosts/"+hostName, 2)
-		r.URL.Path = split[1]
+func (f *HostForwarder) buildProxy(zone, host string) (*httputil.ReverseProxy, error) {
```
ser-io, line 142: nit: buildReverseProxy, keeping the reverse term.

ser-io (approved):

pkg/app/controller.go
jemoreira: Done

pkg/app/types.go
```diff
@@ -61,6 +56,21 @@ type InstanceManager interface {
...
+type HostClient interface {
+	// Get and Post requests return the HTTP status code or an error.
+	// The response body is parsed into the res output parameter if provided.
+	Get(URLPath, URLQuery string, res *HostResponse) (int, error)
+	Post(URLPath, URLQuery string, bodyJSON any, res *HostResponse) (int, error)
+	GetReverseProxy() *httputil.ReverseProxy
+}
+
+type HostResponse struct {
+	responseObj   any
```
ser-io: nit, let's make these fileds public, also remove `response` prefix which is redundant.

ser-io (approved):

pkg/app/types.go
jemoreira: Done

PR #142, opened 2023-04-20
Make host instances visible to `acloud CLI` if AcloudCompatible is true.

ser-io:


pkg/app/gcp/instancemanager_test.go
```diff
@@ -165,8 +167,7 @@ func TestCreateHostRequestBody(t *testing.T) {
     }
   ],
   "labels": {
-    "cf-created_by": "johndoe",
-    "created_by": "cvdremote-johndoe"
+    "cf-created_by": "johndoe"
```
jemoreira, line 170: fakeUsername?

jemoreira (approved):

pkg/app/gcp/instancemanager_test.go
ser-io: done

PR #144, opened 2023-05-08
Add OAuth flow to Cloud Orchestrator

jemoreira:
The OAuth flow is implemented on both local and GCP deployments, however the acquired tokens are not stored in the GCP deployments and are not used for build API access on either deployment yet. Those features will be added later once the flow is tested in QA and production.


kwstephenkim (approved):

conf.toml
```diff
@@ -2,6 +2,16 @@ WebStaticFilesPath = "web"
 
 [AccountManager]
 Type = "unix"
+OAuthRedirectURI = "http://localhost:8080/oauth2callback"
```
Databean, line 5: Assuming this default is for testing only and this will be overridden by most users.

pkg/app/net/googleoauth.go
```diff
@@ -0,0 +1,156 @@
...
+
+func (c *Credentials) AccessToken() string {
+	return c.AccessTk
+}
+
+func (c *Credentials) Expired() bool {
+	return c.expiration.Before(time.Now())
+}
+
+type GoogleOAuthHelper struct {
```
Databean, line 56: Is it possible to use the [Go oauth2 client library](https://github.com/googleapis/google-api-go-client/blob/main/GettingStarted.md) here? The [official Google APIs OAuth page](https://developers.google.com/identity/protocols/oauth2/web-server) doesn't have code examples for it inline but it does mention it at the top.

pkg/app/net/unix/accountmanager.go
```diff
@@ -41,3 +83,12 @@ func (i *UserInfo) Username() string {
 	}
 	return i.username
 }
+
+func oauthCallbackURL(host string) string {
+	return "http://" + host + "/oauth2callback"
```
Databean, line 88: Should this be https?

pkg/app/net/unix/accountmanager.go
```diff
@@ -41,3 +83,12 @@ func (i *UserInfo) Username() string {
...
+}
+
+func randomString() string {
```
Databean, line 91: nit: `randomHexString`?

conf.toml
jemoreira: Yes.
The "unix" stuff is for running the binary locally during development. I decided to include a real url here instead of empty string to make it easier to figure out what to put as redirect URI in the web client configuration in the OAuth server.

pkg/app/net/googleoauth.go
jemoreira: Unless I missed something (wouldn't be the first time) those libraries are for accessing the Google APIs, but don't have (convenient) utilities to manage the oauth credentials. In this case we need the latter and the former is left to `fetch_cvd`.
Even if those libraries do support our use case I couldn't find any good documentation and examples, so it was just easier to use http calls directly.

pkg/app/net/unix/accountmanager.go
jemoreira: The server in the main.go file uses http only (https is added through a reverse proxy when deployed to app engine). So it will be http when running locally. The OAuth spec requires https always except when the host is local (localhost, 127.1 or ::1)

pkg/app/net/unix/accountmanager.go
jemoreira: Done

pkg/app/net/googleoauth.go
Databean: From the [docs](https://pkg.go.dev/golang.org/x/oauth2), which includes a [code sample](https://go.dev/play/p/4VQcYjHb6tG):

- `GoogleOAuthHelper` is [`oauth2.Config`](https://pkg.go.dev/golang.org/x/oauth2#Config)
- `RawCredentials` is [`oauth2.Token`](https://pkg.go.dev/golang.org/x/oauth2#Token)
- `AuthorizationURL` is [`AuthCodeURL`](https://pkg.go.dev/golang.org/x/oauth2#Config.AuthCodeURL)
- `ParseAuthorizationResponse` is [`Exchange`](https://pkg.go.dev/golang.org/x/oauth2#Config.Exchange)
- `RefreshCredentials` maps to either [`TokenSource`](https://pkg.go.dev/golang.org/x/oauth2#Config.TokenSource) or [`Client`](https://pkg.go.dev/golang.org/x/oauth2#Config.Client)

Databean (approved):

pkg/app/net/googleoauth.go
Databean: Sorry, I should have lead with a link to the [oauth2 go package](https://pkg.go.dev/golang.org/x/oauth2) which is referenced there, but is more relevant than the `google-api-go-client` package.

jemoreira: Thanks! Rewrote it with that library instead.

Databean (approved):

PR #147, opened 2023-05-19
persist build api credentials

jemoreira:


conf.toml
```diff
@@ -16,6 +16,15 @@ OAuthClientResourceID = ""
 [SecretManager.UNIX]
 SecretFilePath = "../secrets.json"
 
+[EncryptionService]
+Type = "Simple"
+
+[EncryptionService.Simple]
+KeySizeBits = 128
+
+[EncryptionService.GCP_KMS]
```
ser-io, line 25: we declared above the type to be  "Simple", is this config declaration needed?

pkg/app/net/gcp/spanner.go
```diff
@@ -0,0 +1,81 @@
...
+	"google.golang.org/grpc/codes"
+)
+
+const (
+	credentialsTable  = "Credentials"
+	usernameColumn    = "username"
+	credentialsColumn = "credentials"
+)
+
+// A database service that works with a Cloud Spanner database with the following schema:
```
ser-io, line 42: Let's create a bug to track setting up the database programmatically where we can check in this schema code.

conf.toml
jemoreira: No, I just included it as a reference

pkg/app/types.go
```diff
@@ -108,3 +108,11 @@ type EncryptionService interface {
 	Encrypt(plaintext []byte) ([]byte, error)
 	Decrypt(ciphertext []byte) ([]byte, error)
 }
+
+type DatabaseService interface {
+	// Credentials are usually stored encrypted hence the []byte type.
+	// If no credentials are available for the given user Fetch returns nil, nil.
+	FetchBuildAPICredentials(username string) ([]byte, error)
+	// Store new credentials or overwrite existing ones for the given user.
+	StoreBuildAPICredentials(username string, credentials []byte) error
```
ser-io, line 125: nit: InsertOrUpdate instead of Store?

pkg/app/types.go
```diff
@@ -109,10 +112,21 @@ type EncryptionService interface {
...
+}
+
 type DatabaseService interface {
 	// Credentials are usually stored encrypted hence the []byte type.
 	// If no credentials are available for the given user Fetch returns nil, nil.
 	FetchBuildAPICredentials(username string) ([]byte, error)
 	// Store new credentials or overwrite existing ones for the given user.
 	StoreBuildAPICredentials(username string, credentials []byte) error
+	// Create or update a user session.
+	CreateOrUpdateSession(s *Session) error
```
ser-io, line 127: ideally this object would be immutable, as it's not the case wondering whether we could receive a copy instead so we don't risk the object getting modified before being stored.

conf.toml
ser-io: could it be commented out then? as it gives the impression it's required.

jemoreira: The same applies to the `InstanceManager.GCP` entry below. So far we have put all values, required or not, in this file to make them easier to discover. I'm ok with commenting out the ones that are not required or have default values, but it should be done in a different PR.

ser-io: I see, didn't realize we were following this pattern, discard the comment then.

pkg/app/types.go
jemoreira: Done

ser-io (approved):

PR #151, opened 2023-05-26
Print state changes when creating cvd.

ser-io:


pkg/cli/cli.go
```diff
@@ -910,3 +925,52 @@ func buildServiceRootEndpoint(serviceURL, zone string) string {
...
+	visualsOff := false
+	if f, ok := out.(*os.File); !ok {
+		visualsOff = true
+	} else if !term.IsTerminal(int(f.Fd())) {
+		visualsOff = true
+	}
+	if verbose {
+		visualsOff = true
+	}
+	return &statePrinter{Out: out, Verbose: verbose, visualsOff: visualsOff}
```
jemoreira, line 959: `Verbose` seems to be used only to turn visuals off. If that the case there is no need to store it in the struct.

pkg/cli/cli.go
```diff
@@ -910,3 +925,52 @@ func buildServiceRootEndpoint(serviceURL, zone string) string {
...
+}
+
+func (p *statePrinter) Print(o stateOutput) {
+	format := "\r%s"
```
jemoreira, line 963: IIUC this will cause each line to write over the previous one, but not clear it. So if the second line is shorter that the first one there will be remnants of the first one after printing the second one. In this case I think the second line will look like
```
Creating CVD ....
```
instead of
```
Creating CVD ...
```
because the message is one character shorter. Could you please verify this?

pkg/cli/cli.go
```diff
@@ -910,3 +925,52 @@ func buildServiceRootEndpoint(serviceURL, zone string) string {
 	const version = "v1"
 	return client.BuildRootEndpoint(serviceURL, version, zone)
 }
+
+type stateOutput struct {
+	// State description.
+	Message string
+	// True if state is completed.
+	Done bool
```
jemoreira, line 933: nit: Consider having a `done` message instead of a flag in the state output. It could be something like `doneStateMsg = "\n"

Alternatively, add another method to `statePrinter` to print the "done" marker.

Both approaches avoid having to print the same message twice in lines 577 and 579.

jemoreira (changes requested):

pkg/cli/cli.go
```diff
@@ -910,3 +925,52 @@ func buildServiceRootEndpoint(serviceURL, zone string) string {
...
+//
+// Only use this printer to print user friendly state changes, this is not a logger.
+type statePrinter struct {
+	// Writer to print the messages to. Visual features won't be displayed if not linked to an interactive terminal.
+	Out io.Writer
+	// If true, visual features won't be displayed.
+	Verbose bool
+
+	// If true, visual features like colors and animations won't be displayed.
+	visualsOff bool
```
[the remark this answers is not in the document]

pkg/cli/cli.go
```diff
@@ -910,3 +925,52 @@ func buildServiceRootEndpoint(serviceURL, zone string) string {
...
+}
+
+func newStatePrinter(out io.Writer, verbose bool) *statePrinter {
+	visualsOff := false
+	if f, ok := out.(*os.File); !ok {
+		visualsOff = true
+	} else if !term.IsTerminal(int(f.Fd())) {
+		visualsOff = true
+	}
```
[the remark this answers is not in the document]
ser-io: neat!

pkg/cli/cli.go
ser-io: agree, removed.

pkg/cli/cli.go
ser-io: good catch, used cursor movement character `\033[K` (Erase to end of line). done

pkg/cli/cli.go
ser-io: either in verbose logging runs or while not on an interactive console it's better two see the two different state entries. added a new PrintDone method to the API for better usability.

pkg/cli/cli.go
```diff
@@ -910,3 +925,53 @@ func buildServiceRootEndpoint(serviceURL, zone string) string {
...
+func (p *statePrinter) PrintDone(msg string) {
+	p.print(statePrinterEntry{Message: msg, Done: true})
+}
+
+type statePrinterEntry struct {
+	// State description.
+	Message string
+	// True if state is completed.
+	Done bool
+}
```
jemoreira, line 961: nit: this seems unnecessary now, just passing two parameters to print is fine

jemoreira (approved):

pkg/cli/cli.go

PR #152, opened 2023-05-30
Fix session cookie for oauth

jemoreira:


pkg/app/controller.go
```diff
@@ -327,11 +327,11 @@ func (c *Controller) OAuth2Callback(w http.ResponseWriter, r *http.Request) erro
 	}
 	tk, err := c.oauthConfig.Exchange(oauth2.NoContext, authCode)
 	if err != nil {
-		return err
+		return fmt.Errorf("Error Exchanging token: %w", err)
```
ser-io, line 330: nit: lowercase e.

ser-io (approved):

pkg/app/controller.go
jemoreira: done

PR #161, opened 2023-06-15
Make OAuth2 mandatory for access to build api

jemoreira:


pkg/app/app.go
```diff
@@ -139,17 +139,19 @@ func (a *App) injectCredentialsIntoCreateCVDRequest(r *http.Request, user accoun
...
 		// Get the credentials.
 		tk, err := a.fetchUserCredentials(user)
 		if err != nil {
 			return err
 		}
 		if tk == nil {
 			// There are no credentials in database associated with this user.
-			// TODO(jemoreira) return unauthorized instead.
-			return nil
+			return apperr.NewUnauthenticatedError(
```
ser-io, line 153: We have this in the comments of `NewUnauthenticatedError`:
```
// 401 Unauthorized semantically means "Unauthenticated, while authorization errors are to be
// returned with 403 Forbiden according to
// https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses.
```
should we return 403 instead here as the user is already authenticated.

jemoreira: We don't have the user's credentials, they may or may not be authorized to access the build api we just don't know, so forbidden is not the best reply.

The other way to look at it is: We know the build api is going to return "unauthenticated", so let's go ahead and return it directly ourselves.

Finally, when we implement openid connect, if the user hasn't gone through the oauth2 flow they will truly be unauthenticated. Returning this error here makes handling the response in the client consistent: unauthenticated means go to the auth page and login.

ser-io: I see, however the controller path for this code is protected with the c.Authenticate handler verifying the credentials for the cloud orchestrator service:

`router.Handle("/v1/zones/{zone}/hosts/{host}/{hostPath:.*}", c.Authenticate(c.ForwardToHost))`

from a  code perspective, returning this error here seems inconsistent. Let's add documentation adding this context.

jemoreira: Done

ser-io (approved):

PR #163, opened 2023-06-27
[cli] cvdr pull

ser-io:
- Download runtime artifacts bundle.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -218,7 +218,7 @@ const (
 	AllowAll SelectionOption = 1 << iota
 )
 
-// The PromptSelectionFrom<Type> functions iterate over given container and present
+// The PromptSelectionFromSlice<Type> functions iterate over given container and present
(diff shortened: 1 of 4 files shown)


pkg/cli/cli.go
```diff
@@ -458,7 +464,15 @@ func cvdCommands(opts *subCommandOpts) []*cobra.Command {
 		},
 	}
 	list.Flags().StringVar(&listFlags.Host, hostFlag, "", "Specifies the host")
-	return []*cobra.Command{create, list}
+	// Pull command
+	pull := &cobra.Command{
+		Use:   "pull [HOST]",
```
jemoreira, line 469: Does this mean that it can pull everything from a host, but not a single device when a host has several?

ser-io: yes, it would be based on the output of ` cvd host_bugreport` sub-command which is not device specific for now.

jemoreira (approved):

PR #167, opened 2023-07-10
Show error in case cvd-host_package.tar.gz does not exist.

ser-io:
- The cvd host package tarball is required when creating local image instances.

### DIFF
--- a/pkg/cli/cvd.go
+++ b/pkg/cli/cvd.go
@@ -15,8 +15,10 @@
 package cli
 
 import (
+	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/google/cloud-android-orchestration/pkg/client"
@@ -106,6 +108,10 @@ func (c *cvdCreator) createCVDFromLocalBuild() ([]*hoapi.CVD, error) {
 	if err != nil {
 		return nil, fmt.Errorf("Error building list of required image files: %w", err)
 	}
+	if err := verifyCVDHostPackageTar(vars.HostOut); err != nil {
+		return nil, fmt.Errorf("Invalid cvd host package: %w", err)
+	}
+	names = append(names, filepath.Join(vars.HostOut, CVDHostPackageName))
 	uploadDir, err := c.Service.CreateUpload(c.Opts.Host)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


pkg/cli/cvd.go
```diff
@@ -106,6 +115,21 @@ func (c *cvdCreator) createCVDFromLocalBuild() ([]*hoapi.CVD, error) {
 	if err != nil {
 		return nil, fmt.Errorf("Error building list of required image files: %w", err)
 	}
+	tarName, err := createCVDHostPackageTar(filepath.Join(vars.HostOut, CVDHostPackageDirName))
```
jemoreira, line 118: The cvd-host_package.tar.gz is not supposed to include everything under $ANDROID_HOST_OUT. It would be best to get the tar file created by the build system and ask the user to run `m hosttar` if it doesn't exist.

ser-io: good call, done.

pkg/cli/cvd.go
```diff
@@ -106,6 +108,11 @@ func (c *cvdCreator) createCVDFromLocalBuild() ([]*hoapi.CVD, error) {
 	if err != nil {
 		return nil, fmt.Errorf("Error building list of required image files: %w", err)
 	}
+	cvdHostPkgFilename := filepath.Join(vars.HostOut, CVDHostPackageName)
+	if _, err := os.Stat(cvdHostPkgFilename); errors.Is(err, os.ErrNotExist) {
+		return nil, fmt.Errorf("%q not found. Please run `m hosttar`.", CVDHostPackageName)
```
Databean, line 113: Is there a risk that this file is out of date if the user is required to run `m hosttar` separately? This will remind them to run `m hosttar` the first time, but the user could run `m` later and not `m hosttar`.

jemoreira: could `cvdr` run it itself?

ser-io: Good catch Cody, thanks.

Added an "up to date" logic comparing against `cvd-host_package` directory modification time. PTAL.

Gave it a try to run `m hosttar` programatically and realized that `m` and ` _trigger_build` are shell functions for which didn't find an easy way to propagate, https://docstore.mik.ua/orelly/unix3/upt/ch29_13.htm is the closer I found.

Databean (approved):

jemoreira (approved):

PR #174, opened 2023-07-31
Add CORS origin allow config

k311093:
Add RunningMode field at conf.toml and add CORS handling when mode is in qa mode.


conf.toml
```diff
@@ -1,4 +1,5 @@
 WebStaticFilesPath = "web"
+RunningMode = "prod"
```
ser-io, line 2: we call this concept `environment`. however, using a general `environment` config parameter will result in code like next:

```
if (environment == "qa) {
 // new configs will be set here, from now on
}  
else {
 // or here.
}
``` 

which I think is not good pattern as it makes irrelevant all the existing and new config parameters as you can set them in the `if ... else ...` clause mentioned above. We should rather use an specific config parameters and set the relevant value in each deployment configuration, for example,  in this case we could have something like "EnableCorsException", which could be set true in the config for QA deployment only.

ser-io (changes requested):

k311093:
Changed the config to WebRTC.CorsAllowedOrigins that represents the regex allow list for CORS origins.

Also changed logic for the config WebRTC.CorsAllowedOrigins.

conf.toml
```diff
@@ -43,3 +43,7 @@ HostOrchestratorPort = 1081
 
 [WebRTC]
 STUNServers = ["stun:stun.l.google.com:19302"]
+
+# Regex patterns of allowed origins
+# e.g. "https://localhost:[0-9]+"
+CorsAllowedOrigins = []
```
ser-io, line 49: Let's use `CORSAllowedOrigins` instead.  https://google.github.io/styleguide/go/decisions.html#initialisms

ser-io, line 49: Let's move the config parameter to the general section to the top below `WebStaticFilesPath` as this one is not a `WebRTC` specific configuration paremeter. In the future, we might add a new config group to list general web related config concepts.

conf.toml
```diff
@@ -43,3 +43,7 @@ HostOrchestratorPort = 1081
 
 [WebRTC]
 STUNServers = ["stun:stun.l.google.com:19302"]
+
+# Regex patterns of allowed origins
```
ser-io, line 47: Having regexes here is too permissive and could become a security issue. 

As we are only tackling CORS from **localhost** we could have an specific **`CORSAllowedLocalhost`** boolean config parameter that allows localhost only. On the other hand, if we need to allow external origins in the future we should list them fully rather than through regex matching.

conf.toml
k311093: Changed to use specific config CORSAllowedOrigins instead of depending on environment name.

conf.toml
k311093: Changed to CORSAllowedOrigins.

conf.toml
k311093: Changed to exact match.

conf.toml
k311093: Moved CORSAllowedOrigins config from WebRTC section to the top level.

pkg/app/app.go
```diff
@@ -68,8 +69,24 @@ func NewApp(
...
 	webStaticFilesPath string,
+	corsAllowedOrigins []string,
 	webRTCConfig config.WebRTCConfig) *App {
-	return &App{im, am, oc, es, dbs, webStaticFilesPath, buildInfraCfg(webRTCConfig.STUNServers)}
+	return &App{im, am, oc, es, dbs, webStaticFilesPath, corsAllowedOrigins, buildInfraCfg(webRTCConfig.STUNServers)}
+}
+
+func (c *App) AddCorsHeaderIfNeeded(w http.ResponseWriter, r *http.Request) {
+	origin := r.Header.Get("Origin")
+
```
ser-io, line 79: nit: remove the emtpy lines: 79, 83, 124, 127.

ser-io (approved):
LGTM

pkg/app/app.go
k311093: Removed Removed unnecessary lines.

PR #177, opened 2023-08-02
Use inject build api credentials header in create cvd requests.

ser-io:


pkg/client/client.go
```diff
@@ -323,9 +323,14 @@ func asWebRTCICEServers(in []apiv1.IceServer) []webrtc.ICEServer {
 	return out
 }
 
+const headerNameCOInjectBuildAPICreds = "X-Cutf-Cloud-Orchestrator-Inject-BuildAPI-Creds"
+
 func (c *serviceImpl) CreateCVD(host string, req *hoapi.CreateCVDRequest) (*hoapi.CreateCVDResponse, error) {
+	reqOpts := requestOpts{
+		Header: http.Header{headerNameCOInjectBuildAPICreds: []string{""}},
```
jemoreira, line 330: nit: write something into that string like `"inject"` to make it more intuitive.

jemoreira (approved):

pkg/client/client.go
ser-io: I see, added a comment instead mentioning is fine to use an empty string. I think using a value would make it think that such specific value is required.

PR #181, opened 2023-08-11
Flask mock server for development

retroinspect:
- Mock server imitates behavior of Cloud orchestrator to develop Page 0
  - Includes proxy APIs to Host orchestrator 
- `python server.py` serves APIs under `http://localhost:8071/` in default

### DIFF
--- a/web/page0/.gitignore
+++ b/web/page0/.gitignore
@@ -40,3 +40,6 @@ testem.log
 # System files
 .DS_Store
 Thumbs.db
+
+
+__pycache__
--- a/web/page0/.gitignore
+++ b/web/page0/.gitignore
@@ -42,4 +42,5 @@ testem.log
 Thumbs.db
 
 
+# Python cache files
 __pycache__
--- a/web/page0/src/test_apis.py
+++ b/web/page0/src/test_apis.py
@@ -1,0 +1,10 @@
+"""APIs for test server.
+"""
+import flask
+
+apis = flask.Blueprint('test_apis', __name__)
+
+
+@apis.route('/test')
+def testaaa():
+  return 'this is test'
--- a/web/page0/src/test_apis.py
+++ b/web/page0/src/test_apis.py
@@ -8,6 +8,7 @@ import random
 import string
 import asyncio
 
+
 def get_loop():
     try:
         loop = asyncio.get_running_loop()
@@ -16,6 +17,7 @@ def get_loop():
         return None
     except:
         return None
+
 
 def bg_run(func, second):
     async def task():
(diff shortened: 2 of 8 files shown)


web/page0/src/test_apis.py
```diff
@@ -0,0 +1,330 @@
...
+
+def init():
+    for zone in data["zones"]:
+        mem["zones"][zone] = data["zones"][zone]
+
+
+# GET /info
+@apis.route("/api/info", methods=["GET"])
+def info():
+    sleep(0.1)
```
k311093, line 201: Does this sleep call is need? If this is for simulating the delay of server, delays by network would be sufficient.

web/page0/src/test_apis.py
```diff
@@ -0,0 +1,330 @@
...
+    if loop:
+        loop.create_task(task())
+    else:
+        asyncio.run(task())
+
+apis = flask.Blueprint("test_apis", __name__)
+
+mem = {"info": {"type": "cloud"}, "zones": defaultdict(list)}
+
+default_build_source = {
```
k311093, line 35: If this variable is constant, it would be nice to have a name with all caps like DEFAULT_BUILD_SOURCE

web/page0/.gitignore
```diff
@@ -40,3 +40,6 @@ testem.log
 # System files
 .DS_Store
 Thumbs.db
+
+
```
k311093, line 44: How about adding comment about __pycache__ like other parts of the file? And in this file, it tends to have a space between two sections.

e.g.

# Python cache files

web/page0/src/test_apis.py
```diff
@@ -0,0 +1,330 @@
...
+
+@apis.route("/api/reset", methods=["GET"])
+def reset():
+    init()
+
+
+"""
+Main
+"""
+init()
```
k311093, line 330: If here is Main(), can you please write a code like below?

if __name__ == "__main__":
&nbsp;&nbsp;&nbsp;&nbsp;init()

k311093 (changes requested):

web/page0/src/test_apis.py
retroinspect: Checked that network delay is enough. (>200ms)
Removed all sleep calls.

web/page0/.gitignore
retroinspect: Done

web/page0/src/test_apis.py
retroinspect: Done

PR #185, opened 2023-08-14
Add missing APIs to retrieve cloud orchestration info for Page 0

retroinspect:
- [x] `GET /info` to check validity of the cloud orchestrator & give additional information on the cloud orchestrator
- [x] `GET /v1/zones` to list zones available for GCP instance manager

** Update **
- `GET /v1/info` became `GET /v1/config` to retreive configuration of the cloud orchestrator
- 'Runtime' concept lies on Page 0 UI and not will be revealed on cloud orchestrator side

### DIFF
--- a/api/v1/instancemanager.go
+++ b/api/v1/instancemanager.go
@@ -5,6 +5,10 @@ type CreateHostRequest struct {
 	HostInstance *HostInstance `json:"host_instance"`
 }
 
+type Zone struct {
+	Name string `json:"name"`
+}
+
 type HostInstance struct {
 	// [Output Only] Instance name.
 	Name string `json:"name,omitempty"`
@@ -41,6 +45,10 @@ type OperationResult struct {
 	// method is standard: `Get`/`Create`/`Update`, the response should be the relevant resource
 	// encoded in JSON format.
 	Response string `json:"response,omitempty"`
+}
+
+type ListZonesResponse struct {
+	Items []*Zone `json:"items"`
 }
 
 type ListHostsResponse struct {
--- a/api/v1/instancemanager.go
+++ b/api/v1/instancemanager.go
@@ -60,3 +60,8 @@ type ListHostsResponse struct {
 	// paging through out all the results.
 	NextPageToken string `json:"nextPageToken,omitempty"`
 }
+
+// To be seperated in to new file if the config needs to contain intormation other than instance manager
+type Config struct {
+	InstanceManagerType string `json:"instance_manager_type"`
+}
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -92,6 +92,7 @@ func (c *App) Handler() http.Handler {
 	router := mux.NewRouter()
 
 	// Instance Manager Routes
+	router.Handle("/v1/zones", c.Authenticate(c.listZones)).Methods("GET")
 	router.Handle("/v1/zones/{zone}/hosts", c.Authenticate(c.createHost)).Methods("POST")
 	router.Handle("/v1/zones/{zone}/hosts", c.Authenticate(c.listHosts)).Methods("GET")
 	// Waits for the specified operation to be DONE or for the request to approach the specified deadline,
@@ -182,6 +183,15 @@ func (a *App) injectBuildAPICredsIntoRequest(r *http.Request, user accounts.User
 			"The user must authorize the system to access the Build API on their behalf", nil)
 	}
 	r.Header.Set(headerNameHOBuildAPICreds, tk.AccessToken)
+	return nil
+}
+
+func (c *App) listZones(w http.ResponseWriter, r *http.Request, user accounts.User) error {
+	res, err := c.instanceManager.ListZones()
+	if err != nil {
+		return err
+	}
+	replyJSON(w, res, http.StatusOK)
 	return nil
 }
 
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -58,6 +58,7 @@ type App struct {
 	connectorStaticFilesPath string
 	corsAllowedOrigins       []string
 	infraConfig              apiv1.InfraConfig
+	config                   *config.Config
 }
 
 func NewApp(
@@ -68,8 +69,9 @@ func NewApp(
 	dbs database.Service,
 	webStaticFilesPath string,
 	corsAllowedOrigins []string,
-	webRTCConfig config.WebRTCConfig) *App {
-	return &App{im, am, oc, es, dbs, webStaticFilesPath, corsAllowedOrigins, buildInfraCfg(webRTCConfig.STUNServers)}
+	webRTCConfig config.WebRTCConfig,
+	config *config.Config) *App {
+	return &App{im, am, oc, es, dbs, webStaticFilesPath, corsAllowedOrigins, buildInfraCfg(webRTCConfig.STUNServers), config}
 }
 
 func (c *App) AddCorsHeaderIfNeeded(w http.ResponseWriter, r *http.Request) {
@@ -118,6 +120,7 @@ func (c *App) Handler() http.Handler {
 	router.Handle("/oauth2callback", HTTPHandler(c.OAuth2Callback))
 	router.Handle("/deauth", c.Authenticate(c.DeAuthHandler)).Methods("GET")
 	router.Handle("/deauth", c.Authenticate(c.RescindAuthorizationHandler)).Methods("POST")
+	router.Handle("/v1/config", c.Authenticate(c.ConfigHandler)).Methods("GET")
 	router.Handle("/", c.Authenticate(indexHandler))
 
 	rootRouter := mux.NewRouter()
@@ -382,6 +385,15 @@ func (a *App) RescindAuthorizationHandler(w http.ResponseWriter, r *http.Request
 		return err
 	}
 	fmt.Fprintln(w, "Authorization rescinded")
+	return nil
+}
+
+func (a *App) ConfigHandler(w http.ResponseWriter, r *http.Request, user accounts.User) error {
+	res := apiv1.Config{
+		InstanceManagerType: string(a.config.InstanceManager.Type),
+	}
+
+	replyJSON(w, res, http.StatusOK)
 	return nil
 }
 
--- a/pkg/app/instances/gce.go
+++ b/pkg/app/instances/gce.go
@@ -61,6 +61,26 @@ func NewGCEInstanceManager(cfg Config, service *compute.Service, nameGenerator N
 	}
 }
 
+func (m *GCEInstanceManager) ListZones() (*apiv1.ListZonesResponse, error) {
+	res, err := m.Service.Zones.List(m.Config.GCP.ProjectID).Context(context.TODO()).Do()
+
+	if err != nil {
+		return nil, toAppError(err)
+	}
+
+	var items []*apiv1.Zone
+	for _, i := range res.Items {
+		hi, err := BuildZone(i)
+		if err != nil {
+			return nil, err
+		}
+		items = append(items, hi)
+	}
+	return &apiv1.ListZonesResponse{
+		Items: items,
+	}, nil
+}
+
 func (m *GCEInstanceManager) GetHostAddr(zone string, host string) (string, error) {
 	instance, err := m.getHostInstance(zone, host)
 	if err != nil {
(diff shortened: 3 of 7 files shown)


retroinspect:
@k311093 PTAL

retroinspect:
@ikicha PTAL

pkg/app/instances/gce.go
```diff
@@ -61,6 +61,26 @@ func NewGCEInstanceManager(cfg Config, service *compute.Service, nameGenerator N
...
 
+func (m *GCEInstanceManager) ListZones() (*apiv1.ListZonesResponse, error) {
+	res, err := m.Service.Zones.List(m.Config.GCP.ProjectID).Context(context.TODO()).Do()
+
+	if err != nil {
+		return nil, toAppError(err)
+	}
+
+	var items []*apiv1.Zone
+	for _, i := range res.Items {
```
k311093, line 72: It might be good to change to `for _, i := ...` to `for _, item :=` to prevent confuse with i for index.

pkg/app/instances/gce_test.go
```diff
@@ -53,6 +53,28 @@ func (i *TestUser) Username() string {
...
+
+	if err != nil {
+		t.Errorf("expected <<nil>>, got %+v", err)
+	}
+
+	expectedType := "cloud"
+
+	if resp.RuntimeType != expectedType {
+		t.Errorf("expected <<%q>>, got %q", expectedType, resp.RuntimeType)
+
```
k311093, line 74: nit: remove empty line

pkg/app/instances/gce.go
```diff
@@ -61,6 +61,26 @@ func NewGCEInstanceManager(cfg Config, service *compute.Service, nameGenerator N
...
+		hi, err := BuildZone(i)
+		if err != nil {
+			return nil, err
+		}
+		items = append(items, hi)
```
k311093, line 77: If there is no possibility of errors while converting compute.Zone to apiv1.Zone, how about changing these lines to 

```
items = append(items, &apiv1.Zone{ Name: item.Name })
```

and remove BuildZone() function.

k311093 (changes requested):

pkg/app/instances/gce_test.go
retroinspect: done

pkg/app/instances/gce.go
retroinspect: done

k311093:
Looks good to me.

@jemoreira , @ser-io can you please take a look?

api/v1/instancemanager.go
```diff
@@ -52,3 +52,7 @@ type ListHostsResponse struct {
 	// paging through out all the results.
 	NextPageToken string `json:"nextPageToken,omitempty"`
 }
+
+type Info struct {
```
ser-io, line 64: `Info` is quite generic in this context. From the implementation it seems this is specific to the Instance Manager configuration, perhaps `InstanceManagerConfig`. Let's add documentation to this `struct` and its fields as well.

api/v1/instancemanager.go
```diff
@@ -52,3 +52,7 @@ type ListHostsResponse struct {
...
+	RuntimeType string `json:"type"`
```
ser-io, line 65: If using `InstanceManagerConfig`, this filed could be  just `Type` instead.

pkg/app/app.go
```diff
@@ -116,6 +116,7 @@ func (c *App) Handler() http.Handler {
 	router.Handle("/oauth2callback", HTTPHandler(c.OAuth2Callback))
 	router.Handle("/deauth", c.Authenticate(c.DeAuthHandler)).Methods("GET")
 	router.Handle("/deauth", c.Authenticate(c.RescindAuthorizationHandler)).Methods("POST")
+	router.Handle("/info", c.Authenticate(c.InfoHandler)).Methods("GET")
```
ser-io, line 120: let's use the `/v1` prefix here as this logic  belongs to the REST API.

pkg/app/app.go
```diff
@@ -414,6 +415,16 @@ func (a *App) RescindAuthorizationHandler(w http.ResponseWriter, r *http.Request
 	return nil
 }
 
+func (a *App) InfoHandler(w http.ResponseWriter, r *http.Request, user accounts.User) error {
+	res, err := a.instanceManager.GetInfo()
+	if err != nil {
+		return nil
```
ser-io, line 431: `return err`

pkg/app/instances/gce.go
```diff
@@ -347,3 +347,9 @@ sudo -u vsoc-01 touch /home/vsoc-01/bin/restart_cvd
 printf '#!/bin/bash\nHOME=/var/lib/cuttlefish-common/runtimes /var/lib/cuttlefish-common/artifacts/acloud_link/bin/restart_cvd\n' | sudo -u vsoc-01 tee /home/vsoc-01/bin/restart_cvd >/dev/null
 sudo -u vsoc-01 chmod u+x /home/vsoc-01/bin/restart_cvd
 `
+
+func (m *GCEInstanceManager) GetInfo() (*apiv1.Info, error) {
+	return &apiv1.Info{
+		RuntimeType: "cloud",
```
ser-io, line 369: suggest to declare the following values: "cloud", "local" as constants in a common place, it could be in the API definition file, so the clients are aware of what these values look like when reading the public API documentation.

ser-io (changes requested):

pkg/app/instances/gce.go
```diff
@@ -347,3 +363,9 @@ sudo -u vsoc-01 touch /home/vsoc-01/bin/restart_cvd
...
```
jemoreira, line 369: "cloud" is a little ambiguous as it could be AWS, GCP, etc. Even within GCP it could be GCE or GKE.
If you wish to differentiate between "cloud" and "local" or "test", then it would be worth it to add a `name` field to the response and name this one "gce". Or simply have its type be "gce".

api/v1/instancemanager.go
retroinspect: Considered `InstanceManagerConfig` as the more concrete name before, but it may be confused with `Config` in `pkg/app/instances/instances.go`.

I think that runtime is different from an instance manager as runtime is the system running cloud orchestrator, and instance manager is a component of cloud orchestrator.

Plus, `GET /info` was designed to return architecture or OS version which seem to be not configurable. Or is it okay to name Config something we cannot configure?
If so, how about `RuntimeConfig` to distinguish it from `InstanceManagerConfig`

pkg/app/app.go
retroinspect: Done

pkg/app/app.go
retroinspect: Done

pkg/app/instances/gce.go
retroinspect: Done

pkg/app/instances/gce.go
retroinspect: `RuntimeType` was first introduced to just color runtime cards in Page 0 differently based on some type of runtime.

I think that whether the cloud orchestrator runs from GCP/AWS/etc doesn't have to be shown to Page 0 as far.
If such information should be revealed, we can return `InstanceManagerConfig` without defining new type.

api/v1/instancemanager.go
ser-io: Oh, I see where the confusion is at.

There are two different concepts here whose values overlap: `Runtime` where the cloud orchestrator is deployed to, and `InstanceManagerType`, the implementation type used by the `cloud orchestartor` to perform instance management operations.

`Runtime` could have values like: `GAE (Google App Engine)` and `local`, whereas `InstanceManagerType` could have values like: `GCP (Google Cloud Platform)`, `AWS` and `local`. This mean that we could have a `local` deployment and do instance management in GCP for example.

I suggested `InstanceManagerConfig` as we were modifying the `InstanceManager` interface to provide this information. I'm not sure which concept fits your use case better here. In either case I think there's no need to  modify the `InstanceManager` interface as we can use the cloud orchetrator configuration to gather relevant information. The config is loaded here https://github.com/google/cloud-android-orchestration/blob/main/cmd/cloud_orchestrator/main.go#L162.

retroinspect: > In either case I think there's no need to modify the InstanceManager interface as we can use the cloud orchetrator configuration to gather relevant information.

I see, thus made a new file to define the `Info` interface.

ser-io: I think it's fine to add the service information handling logic into the `app.go` file for now.

retroinspect: Since cloud orchestrators with GCP/local instance managers should return different runtime types. is it okay to maintain current function locations?

ser-io: Right now we have something like:

```
func (m *GCEInstanceManager) GetInfo() (*apiv1.Info, error) {
	return &apiv1.Info{
		RuntimeType: apiv1.Cloud,
	}, nil
}
```

and this is not always true as  we could do instance management in GCP using a local deployment. 

Gathering **runtime** information ( information about the platform the `cloud orchestrator` has been deployed to) is common no matter the instance manager type that was in its configuration.

retroinspect: I think I was confused 😅  
Then is it proper to get runtime type from `conf.toml`?  
I think it can be included as part of `type Config`.

retroinspect: @ser-io PTAL

api/v1/instancemanager.go
retroinspect: Updated: use `Config` as API response & `instanceManagerType` as field name

api/v1/instancemanager.go
retroinspect: Updated: now returns `config.Config` as source of data which is parsed result of `conf.toml`

pkg/app/instances/gce.go
retroinspect: Updated: Return `InstanceManagerConfig` as is and **decide the runtime type at Page 0 UI side**  based on instance manager type

retroinspect:
I updated the PR based on discussion with @ser-io .
Please take a look again 🙏  @jemoreira @ser-io

ser-io (changes requested):
@retroinspect good work in [Change GET /v1/info to GET /v1/config](https://github.com/google/cloud-android-orchestration/pull/185/commits/a6afca31079b338d8e2aa47bbb45c409b5695b69)


We have 9 commits in this PR, the ones related to the previous `info` logic (I think 5 of them) are reverted in code in [Change GET /v1/info to GET /v1/config](https://github.com/google/cloud-android-orchestration/pull/185/commits/a6afca31079b338d8e2aa47bbb45c409b5695b69) and not in log history. This will introduce  noise to the main branch history and we can avoid that.

My recommendation is to create a new PR cherry picking the commits associated with ` GET /v1/zones` logic.

Then, in this PR, you can `hard reset` the previous 8 commits and `force-push` with only the last one. Having only  [Change GET /v1/info to GET /v1/config](https://github.com/google/cloud-android-orchestration/pull/185/commits/a6afca31079b338d8e2aa47bbb45c409b5695b69) on this PR.  

WDYT?

retroinspect:
> @retroinspect good work in [Change GET /v1/info to GET /v1/config](https://github.com/google/cloud-android-orchestration/pull/185/commits/a6afca31079b338d8e2aa47bbb45c409b5695b69)
> 
> We have 9 commits in this PR, the ones related to the previous `info` logic (I think 5 of them) are reverted in code in [Change GET /v1/info to GET /v1/config](https://github.com/google/cloud-android-orchestration/pull/185/commits/a6afca31079b338d8e2aa47bbb45c409b5695b69) and not in log history. This will introduce noise to the main branch history when merged and we can avoid that.
> 
> My recommendation is to create a new PR cherry picking the commits associated with ` GET /v1/zones` logic.
> 
> Then, in this PR, you can `hard reset` the previous 8 commits and `force-push` with only the last one. Having only [Change GET /v1/info to GET /v1/config](https://github.com/google/cloud-android-orchestration/pull/185/commits/a6afca31079b338d8e2aa47bbb45c409b5695b69) on this PR.
> 
> WDYT?


I hard reset the previous commits on 'GET /info` and `force-push`ed with only the last one.

However, I think if we separate  `zones` PR and `config` PR, `zones` will depend on `config` as `struct App` requires a new field (`config.Config`).

To merge & review only once, I suggest to use this PR for both `zones` and `config`.
WDYT?

ser-io:
LGTM. Please update the current PR description: `Add missing APIs to retrieve runtime info for Page 0` before merging.

ser-io (approved):

retroinspect:
Changed!

PR #187, opened 2023-08-21
Add additional settings to handle CORS

retroinspect:
- `CorsInterceptor` in frontend & `Access-Control-Allow-Credentials` in cloud orchestrator allows HTTP requests to carry cookies
- Explicitly added allowed methods 
- Handle preflight request as specified from [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS)

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -82,6 +82,9 @@ func (c *App) AddCorsHeaderIfNeeded(w http.ResponseWriter, r *http.Request) {
 	for _, allowed := range c.corsAllowedOrigins {
 		if origin == allowed {
 			w.Header().Add("Access-Control-Allow-Origin", origin)
+			w.Header().Add("Access-Control-Allow-Methods", "GET, POST, OPTIONS, PUT, DELETE")
+			w.Header().Add("Access-Control-Allow-Headers", "Content-Type")
+			w.Header().Add("Access-Control-Allow-Credentials", "true")
 			break
 		}
 	}
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -46,6 +46,7 @@ import (
 
 const (
 	sessionIdCookie = "sessionid"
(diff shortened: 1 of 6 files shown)


pkg/app/app.go
```diff
@@ -122,6 +126,13 @@ func (c *App) Handler() http.Handler {
 	rootRouter.PathPrefix("/").Handler(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		c.AddCorsHeaderIfNeeded(w, r)
 		router.ServeHTTP(w, r)
```
retroinspect, line 128: TODO: should be removed

retroinspect: Done

retroinspect:
@k311093 PTAL

pkg/app/app.go
```diff
@@ -121,7 +125,13 @@ func (c *App) Handler() http.Handler {
 	rootRouter := mux.NewRouter()
 	rootRouter.PathPrefix("/").Handler(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		c.AddCorsHeaderIfNeeded(w, r)
-		router.ServeHTTP(w, r)
+		if r.Method == "OPTIONS" {
+			w.Header().Add("Allow", allowedMethods)
```
k311093, line 129: How about using r.Method instead of allowedMethods since it also allows the methods that is not requested.

k311093 (changes requested):

pkg/app/app.go
retroinspect: According to the [specification](https://httpwg.org/specs/rfc9110.html#field.allow), `Allow` header contains all request methods the server supports.
 While `r.Method` here always results in "OPTIONS", I think `allowedMethods` would be appropriate for the spec.

k311093: Acknowledged, thanks!

PR #194, opened 2023-08-29
[cvdr] Rename `cvd` related structs.

ser-io:
- New names aim to be more specifc.

### DIFF
--- a/pkg/cli/cvd.go
+++ b/pkg/cli/cvd.go
@@ -27,22 +27,22 @@ import (
 	"github.com/hashicorp/go-multierror"
 )
 
-type CVD struct {
+type RemoteCVDLocator struct {
 	ServiceRootEndpoint string `json:"service_root_endpoint"`
 	Host                string `json:"host"`
 	Name                string `json:"name"`
 }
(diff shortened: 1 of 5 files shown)


pkg/cli/cvd.go
```diff
@@ -27,22 +27,22 @@ import (
...
-type CVD struct {
+type RemoteCVDLocator struct {
 	ServiceRootEndpoint string `json:"service_root_endpoint"`
 	Host                string `json:"host"`
 	Name                string `json:"name"`
 }
 
-type CVDInfo struct {
-	CVD
+type RemoteCVD struct {
```
jemoreira, line 36: If we never have "local CVDs" (which I don't think make sense in this context) adding "Remote" to the names is redundant.

jemoreira (approved):

pkg/cli/cvd.go
ser-io: Used the `Remote` prefix to differentiate with the `CVD` API object, I understand these are in different modules but it makes it clearer.

PR #195, opened 2023-08-30
Page0 long polling for hosts and environments

retroinspect:
- Should be merged after #191 
- Add long polling for host creation, host deletion, and env creation
- Host deletion updates corresponding environment to be status "stopping"

### DIFF
--- a/web/page0/src/app/env.service.ts
+++ b/web/page0/src/app/env.service.ts
@@ -1,9 +1,12 @@
 import {Injectable} from '@angular/core';
-import {tap} from 'rxjs/operators';
+import {throwError} from 'rxjs';
+import {catchError, tap} from 'rxjs/operators';
 import {Store} from 'src/app/store/store';
 import {ApiService} from './api.service';
 import {GroupForm} from './interface/device-interface';
 import {Environment, EnvStatus} from './interface/env-interface';
+import {Group} from './interface/host-orchestrator.dto';
+import {OperationService} from './operation.service';
 import {hasDuplicate} from './utils';
 @Injectable({
   providedIn: 'root',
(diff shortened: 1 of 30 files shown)


retroinspect:
@ikicha @k311093 PTAL

k311093:
Can you please rebase this PR to the main?

retroinspect:
> Can you please rebase this PR to the main?

Done!

web/page0/src/app/host.service.ts
```diff
@@ -1,30 +1,102 @@
...
+            wait: {
+              waitUrl,
+              metadata: {
+                type: 'host-create',
+                zone,
+                runtimeAlias: runtime.alias,
+              },
+            },
+          });
+          this.waitService.longPolling<HostInstance>(waitUrl).subscribe({
```
ikicha, line 34: instead of subscribe in tap, you can use switch map for that instead.
tap(operation => this.store.dispatch({...}))
.switchMap(operation => this.waitService.longPolling~~~)?

web/page0/src/app/env.service.ts
```diff
@@ -38,32 +41,67 @@ export class EnvService {
...
+                type: 'env-create',
+                hostUrl,
+                groupName,
+                runtimeAlias,
+                devices,
+              },
+            },
+          });
+
+          this.waitService.longPolling<Group>(waitUrl).subscribe({
```
ikicha, line 60: the same as below

web/page0/src/app/fetch.service.ts
```diff
@@ -118,5 +125,46 @@ export class FetchService {
     );
   }
 
-  constructor(private apiService: ApiService) {}
+  loadHosts(runtime: Runtime) {
+    if (runtime.status !== RuntimeStatus.valid || !runtime.zones) {
+      return of([]);
+    }
+
+    return forkJoin(
```
ikicha, line 133: instead of nested call, how about using merge?

hosts  = runtime.zones.flatmap(~~~);
envs = hosts.flatmap(...)

merge(hosts.map(make_message), env.map(make_message)).subscribe(this.store.dispatch)

return envs.scan or something

web/page0/src/app/host.service.ts
retroinspect: Subscription from tap is needed to meet current requirements:
- On completion of POST /hosts, the user navigates back to previousUrl immediately
- EnvService does not have dependencies on RouterModule or Snackbar as they are sort of UI modules

web/page0/src/app/fetch.service.ts
retroinspect: Refactored.

Cannot subscribe inside of `loadHosts` since all the API calls should be canceled if RefreshService cancels the previous subscription. (i.e. the user presses the refresh button many times)

The resulting Observable of loadHosts is only used to mark completion of all API calls, thus annotated type as Observable<any>.

ikicha (approved):

PR #199, opened 2023-09-04
Page0 host "auto create" option for Create Env view

retroinspect:
- Add "auto create" option in host selector for Create Env view so that users do not have to leave the view to create a new host 
- With the auto create option, the app automatically creates a host with default setting and send POST /cvds to the created host


k311093:
This PR would be followed-up at another PR (https://github.com/google/cloud-android-orchestration/pull/206)

PR #204, opened 2023-10-02
Add metrics package to cvdr

Mahmod:
The initial structure for metrics package that will be extended in the future to implement metrics client for cvdr


ser-io (commented):
let's move the new packe into the `cli` package as it's specific to the cli binary (cvdr).

ser-io:
can the commit `386e6edae019ebab8e48ff640c129b63024fcee7` be removed from this PR as it was already merged in https://github.com/google/cloud-android-orchestration/pull/203?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,38 @@
...
+
+// Mock function to send data to metrics system.
+func sendToMetrics(data string) {
+	// TODO: Replace with actual implementation to send data to your metrics system.
+	println("Sending to metrics:", data)
+}
+
+// Function to extract the launch command.
+func extractLaunchCommand() string {
+	launchCommand := strings.Join(os.Args, " ")
```
ser-io, line 30: nit: `return strings.Join(os.Args, " ")` directly?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,38 @@
...
+	return launchCommand
+}
+
+// Function to send the launch command to metrics.
+func sendLaunchCommandToMetrics() {
+	launchCommand := extractLaunchCommand()
```
ser-io, line 36: nit: `c` or `cmd` are fine here instead of larger `launchCommand` name following Golang naming conventions: https://google.github.io/styleguide/go/decisions#variable-names.

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,28 @@
...
+// metrics/metrics_test.go
+
+package metrics
+
+import (
+	"testing"
+)
+
+func TestGetLaunchCommand(t *testing.T) {
+	command := extractLaunchCommand()
```
jemoreira, line 23: This test isn't useful: The contents of `os.Args` during test execution are completely different than those during execution of the `cvdr` command.
Furthermore, `os.Args` is a non constant global variable and could be changed anywhere in the code. If something in the test framework decides to change that the test could fail.

pkg/metrics/metrics.go
jemoreira, line 36: Instead of extracting the launch command from a global variable that could have been changed elsewhere in the program it's better to accept the command arguments as a parameter to the function. I suspect this function will be called from a place very close to where those arguments are parsed, so it should't be too hard to make sure the arguments passed to this function are the same as those parsed for the command.

jemoreira (changes requested):

Mahmod:
I'll close this pull request , no longer valid

PR #205, opened 2023-10-10
Page0 canonical config

k311093:
From previous PR

- Use JSON canonical config to call POST /cvds
- Canonical config is synchronized with manual form if possible


k311093:
@ikicha Can you please take a look?

ikicha:
Sorry for the late response, LGTM!

PR #208, opened 2023-10-11
Revert "README updates and copyright headers."

ser-io:
Reverts google/cloud-android-orchestration#203

Breaks building the debian package as `goutil` script does no longer start with shebang (#!)


jemoreira:
Sounds like you just need to put the `#!/bin/bash` line at the beginning, not revert the whole thing. Unless you want to submit it again with the fix.

jemoreira (approved):

PR #211, opened 2023-10-13
Adds top level `delete` subcommand.

ser-io:
- There's no deletion at CVD granularity for now, CVD(s) are deleted by deleting its host.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -511,7 +511,15 @@ func cvdCommands(opts *subCommandOpts) []*cobra.Command {
 			return runPullCommand(c, args, opts.RootFlags, opts)
 		},
 	}
-	return []*cobra.Command{create, list, pull}
+	// Delete command
+	del := &cobra.Command{
+		Use:   "delete <foo> <bar> <baz>",
+		Short: "Delete hosts",
+		RunE: func(c *cobra.Command, args []string) error {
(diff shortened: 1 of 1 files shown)


pkg/cli/cli.go
```diff
@@ -511,7 +511,15 @@ func cvdCommands(opts *subCommandOpts) []*cobra.Command {
 			return runPullCommand(c, args, opts.RootFlags, opts)
 		},
 	}
-	return []*cobra.Command{create, list, pull}
+	// Delete command
+	del := &cobra.Command{
+		Use:   "delete <foo> <bar> <baz>",
+		Short: "Delete hosts",
+		RunE: func(c *cobra.Command, args []string) error {
+			return runDeleteHostsCommand(c, args, opts.RootFlags, opts)
```
jemoreira, line 519: This command should "delete" single cvd instances, not entire hosts.

jemoreira (changes requested):

pkg/cli/cli.go
ser-io: my idea was that it could handle both:, i.e

```
cvdr delete

(1) foo/cvd-1
(2) foo/cvd-2
(3) foo(all)

```

however the `foo(all)` option should delete all the `cvds` in the host, and not delete the host itself. I'll abandon this PR.

PR #217, opened 2023-10-18
Import relevant acloud configuration values.

ser-io:


cmd/cvdr/main.go
```diff
@@ -46,7 +46,16 @@ func loadInitialConfig() (*cli.Config, error) {
 				return nil, err
 			}
 		} else if errors.Is(err, os.ErrNotExist) {
-			// TODO: Create a new one importing relevant acloud config parameters.
+			// Load
+			acloudConfigPath := cli.ExpandPath("~/.config/acloud/acloud.config")
+			if _, err := os.Stat(acloudConfigPath); err == nil {
+				if err := cli.ImportAcloudConfig(acloudConfigPath, path); err != nil {
```
jemoreira, line 52: nit: It would be nice obtain user's consent to do this first. Print something like "No user configuration found, would you like to generate it by importing your acloud configuration?".

cmd/cvdr/main.go
```diff
@@ -46,7 +46,16 @@ func loadInitialConfig() (*cli.Config, error) {
 				return nil, err
```
jemoreira, line 46: nit: I would re-order this `if` like this:
```
_, err := os.Stat(path)
if errors.Is(err, os.ErrNotExist) {
  // Import the acloud config
  err = nil
}
if err == nil {
  // Load config file
} else {
  // return error
}
```

That way loading the config file is done in a single place.

pkg/cli/config.go
```diff
@@ -94,3 +96,67 @@ func buildConfig(sources []io.Reader) (*Config, error) {
...
+		return fmt.Errorf("failed encoding: %w", err)
+	}
+	return nil
+}
+
+// Extract relevant acloud config values.
+// Acloud config is written in Proto Text Format https://protobuf.dev/reference/protobuf/textformat-spec/
+func extractAcloudConfig(input string) (AcloudConfig, error) {
+	extract := func(field, input string) (string, error) {
+		re := regexp.MustCompile(`(^|\n)` + field + `:\s+\"([^"]+)\"`)
```
jemoreira, line 142: This regex will miss lines like these:
```
field:"value"
   field: "value"
field : "value"
filed: "val\"ue"
```
A better regex would be `[\n^]\s*` + field + `\s*:\s*\"(.*)\"\s*[\n$]`, note that the value will be in `matches[1]`, not `matches[2]`.

pkg/cli/config_test.go
```diff
@@ -121,6 +121,91 @@ MachineType = "machine-type-bar"
...
+func TestImportAcloudConfig(t *testing.T) {
+	const acloudConfig = `project: "foo" # project zero
+# Zone comment
+zone: "foo-central1-c" # zone: "invalid"
+# machine_type: "comment"
+machine_type: "foo-standard-4"
+network: "default"
+
+# Cuttlefish host image
+# stable_host_image_family: "acloud-release"
```
jemoreira, line 133: you need more test cases, for example spaces and tabs in any unexpected place in the line.

cmd/cvdr/main.go
ser-io: good point, done

cmd/cvdr/main.go
ser-io: good point, done.

pkg/cli/config_test.go
ser-io: done

pkg/cli/config.go
ser-io: good catch, added the whitespaces and used a stricter group matching due the nature of these values.

jemoreira (approved):

PR #222, opened 2023-10-25
Default to yes in yes/no prompt logic.

ser-io:


pkg/cli/cli.go
```diff
@@ -326,9 +326,10 @@ func PromptSelectionFromMap[K comparable, T any](c *command, choices map[K]T, to
 
 func PromptYesOrNo(out *os.File, in *os.File, text string) (bool, error) {
 	fmt.Fprint(out, text+" (y/N): ")
```
jemoreira, line 328: `(Y/n)`

ser-io: Done. Also change the logic to accept lower and upper case values for both: `y` and `n`.

pkg/cli/cli.go
```diff
@@ -325,16 +325,18 @@ func PromptSelectionFromMap[K comparable, T any](c *command, choices map[K]T, to
...
-	return yN == "y", nil
+	fmt.Fprint(out, text+" (Y/n): ")
+	yn := "Y"
+	_, err := fmt.Fscanln(os.Stdin, &yn)
+	// Using the error text for comparison as there's no specific error type to compare against.
+	if err != nil && err.Error() != "unexpected newline" {
+		return false, fmt.Errorf("failed to read (Y/n) choice: %w", err)
+	}
+	ynLo := strings.ToLower(yn)
+	if ynLo != "y" && ynLo != "n" {
```
jemoreira, line 336: nit: the user could write "yes" or "no", include those here and in the return compare `ynLo[0] == 'y'`

jemoreira (approved):

pkg/cli/cli.go
ser-io: Done.

jemoreira (approved):

PR #223, opened 2023-10-26
Make host orchestrator client usable outside of the cloud orchestrator context

jemoreira:


pkg/client/host_orchestrator_client.go
```diff
@@ -39,6 +39,7 @@ type HostOrchestratorService interface {
 
 	// Uploads user files to a previously created directory.
 	UploadFiles(uploadDir string, filenames []string) error
+	UploadFilesWithOptions(uploadDir string, filenames []string, options UploadOptions) error
```
ser-io, line 42: nit: empty line above.

pkg/client/httputils.go
```diff
@@ -210,13 +192,20 @@ type fileInfo struct {
 	TotalChunks int
 }
 
+type UploadOptions struct {
+	InitialDuration     time.Duration
```
ser-io, line 196: let's group these 4 fields using a new struct, i.e `BackOffOpts` as they are meant to be used together.

pkg/client/host_orchestrator_client.go
jemoreira: Done

pkg/client/httputils.go
jemoreira: Done

ser-io (approved):

PR #226, opened 2023-10-31
Improve error message.

ser-io:
- Makes it specific the errors are about the canonical configuration file.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -641,14 +641,15 @@ const (
 
 func runCreateCVDCommand(c *cobra.Command, args []string, flags *CreateCVDFlags, opts *subCommandOpts) error {
 	if len(args) > 0 {
-		// Load and parse the passed environment configuration.
-		data, err := os.ReadFile(args[0])
+		// Load and parse the passed environment specification.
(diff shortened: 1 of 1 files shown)


pkg/cli/cli.go
```diff
@@ -641,14 +641,15 @@ const (
 
 func runCreateCVDCommand(c *cobra.Command, args []string, flags *CreateCVDFlags, opts *subCommandOpts) error {
 	if len(args) > 0 {
-		// Load and parse the passed environment configuration.
-		data, err := os.ReadFile(args[0])
+		// Load and parse the passed canonical configuration.
```
jemoreira, line 644: environment is the right word, canonical are the ones that you get by just importing @phone or @tablet inside those files.

ser-io: I see, changed it to `environment specification` as discussed offline.

jemoreira, line 644: missed the "canonical" in this comment

jemoreira (approved):

pkg/cli/cli.go
ser-io: Done.

PR #234, opened 2023-11-08
Apply staticchecker linter and resolve styling  issues.

Mahmod:
Resolved patterns:
ST1005 {error  messages}
error strings should not be capitalized

[reference](https://google.github.io/styleguide/go/decisions.html#error-strings)

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -265,15 +265,15 @@ func (c *App) OAuth2Callback(w http.ResponseWriter, r *http.Request) error {
 	}
 	tk, err := c.oauth2Helper.Exchange(oauth2.NoContext, authCode)
 	if err != nil {
-		return fmt.Errorf("Error exchanging token: %w", err)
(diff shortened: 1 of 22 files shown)


pkg/app/app.go
```diff
@@ -529,7 +529,7 @@ func (a *App) findInterceptFile(hostPath string) (string, bool) {
 func HostOrchestratorPath(path, host string) (string, error) {
 	split := strings.SplitN(path, "hosts/"+host, 2)
 	if len(split) != 2 {
-		return "", fmt.Errorf("URL path doesn't have host component: %s", path)
+		return "", fmt.Errorf("uRL path doesn't have host component: %s", path)
```
jemoreira, line 532: `url` all in lowercase

jemoreira (changes requested):

pkg/app/app.go
Mahmod: done

PR #238, opened 2023-11-08
Address Leaks and Go Vet Tool Issues

Mahmod:
This commit fixes various leaks and other errors identified by the go vet tool across the Golang codebase.

Please review this pull request carefully , since I only validated the suggested fixes through `go vet ./...` command .
I'm not familiar with the functional testing that assure the functionality of my suggestions .

Bug: b/309808700

### DIFF
--- a/pkg/client/httputils.go
+++ b/pkg/client/httputils.go
@@ -218,6 +218,9 @@ func (u *FilesUploader) Upload(files []string) error {
 		return err
 	}
 	ctx, cancel := context.WithCancel(context.Background())
+	// ensure cancel is called even in the case of an error
+	defer cancel()
+
 	jobsChan := make(chan uploadChunkJob)
 	resultsChan := u.startWorkers(ctx, jobsChan)
 	go func() {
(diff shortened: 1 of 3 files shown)


pkg/client/httputils.go
```diff
@@ -232,11 +235,16 @@ func (u *FilesUploader) Upload(files []string) error {
...
 			}
 		}
 	}
+
+	// Ensure all results are received to prevent goroutine leaks.
+	defer func() {
+		for range resultsChan {
+			// Drain the channel.
+		}
+	}()
```
jemoreira, line 247: This isn't needed. The loop above doesn't break until the channel is drained.

pkg/client/httputils.go
```diff
@@ -232,11 +235,16 @@ func (u *FilesUploader) Upload(files []string) error {
 			if returnErr == nil {
 				returnErr = err
```
jemoreira, line 236: you can improve this line by making `returnErr` a `multierror` and appending each received error to it.

pkg/client/httputils.go
```diff
@@ -218,6 +218,9 @@ func (u *FilesUploader) Upload(files []string) error {
 		return err
 	}
 	ctx, cancel := context.WithCancel(context.Background())
+	// ensure cancel is called even in the case of an error
```
jemoreira, line 221: Actually, the existing code only calls `cancel` when there are errors. This ensures it's called when there are NO errors too.

jemoreira (changes requested):

Mahmod:
This issue is resolved by @jemoreira in different Pr , no longer needed

PR #244, opened 2023-11-17
Make running the cloud and host orchestrators locally easy and convenient.

jemoreira:
After this you can run these commands to test the system end to end:
```
$ cvd_host_orchestrator 2>ho.err >ho.out & # requires cuttlefish-user v0.9.28
$ go build ./cmd/cloud_orchestrator/ && ./cloud-orchestrator &
$ go build ./cmd/cvdr && ./cvdr --service_url=http://localhost:8080 --zone=z create --host=local --credential_source=none 
```

### DIFF
--- a/build/package/host/etc/cvdr.toml
+++ b/build/package/host/etc/cvdr.toml
@@ -25,6 +25,12 @@
 # files forever.
 # KeepLogFilesDays = 30
 
+# Source for the Build API credentials to be used on cvd create operations.
+# Possible values are:
+#  - "none": Don't use Build API credentials.
+#  - "injected": Use credentials stored in the server.
+# BuildAPICredentialsSource = "injected"
+
 # The host.GCP section provides default configuration values for hosts created
 # using the GCP Cloud Provider. Other providers will have their own
 # host.<Provider> sections when added.
--- a/pkg/cli/cvd.go
+++ b/pkg/cli/cvd.go
@@ -62,6 +62,11 @@ func NewRemoteCVD(url, host string, cvd *hoapi.CVD) *RemoteCVD {
 	}
 }
 
+const (
+	NoneCredentialsSource     = "none"
+	InjectedCredentialsSource = "injected"
+)
+
 type CreateCVDOpts struct {
 	Host            string
 	MainBuild       hoapi.AndroidCIBuild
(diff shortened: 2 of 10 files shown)


build/package/host/etc/cvdr.toml
```diff
@@ -25,6 +25,12 @@
 # files forever.
 # KeepLogFilesDays = 30
 
+# Source for the Build API credentials to be used on cvd create operations.
+# Possible values are:
+#  - "injected": Use credentials stored in the server.
+#  - "none": Don't use Build API credentials.
+# CredentialsSource = "injected"
```
ser-io, line 32: should we be more specific here: `BuildAPICredentialsSource`? 

First interpretation of CredentialsSource could be about credentials against the Cloud Orchestrator service itself, I could foresee users referencing this config if authentication against the CO fails.

build/package/host/etc/cvdr.toml
```diff
@@ -25,6 +25,12 @@
 # files forever.
 # KeepLogFilesDays = 30
 
+# Source for the Build API credentials to be used on cvd create operations.
+# Possible values are:
+#  - "injected": Use credentials stored in the server.
+#  - "none": Don't use Build API credentials.
```
ser-io, line 31: should we use `"empty": Use empty string credentials`? From what I gather in the code it seems the "none" concept ends up returning an empty string as the credential value. 

nit: let's also move either "none" or "empty" to the first option as it's the trivial one.

pkg/cli/cvd.go
```diff
@@ -218,14 +235,25 @@ func (c *cvdCreator) createWithOpts() ([]*hoapi.CVD, error) {
...
 }
 
+func credentialsFactoryFromSource(source string) (CredentialsFactory, error) {
+	switch source {
+	case NoneCredentialsSource:
+		return func() string { return "" }, nil
+	case InjectedCredentialsSource:
+		return func() string { return client.InjectedCredentials }, nil
+	default:
+		return nil, fmt.Errorf("unknown credentials source: %s", source)
```
ser-io, line 253: nit: would this be a good place to use `panic` instead?

jemoreira: no: the library shouldn't panic, the executable binary that uses it can panic if it wants to, but not the library.
The rule of thumb is to panic only if the state of the system is corrupted, otherwise use regular error handling.

build/package/host/etc/cvdr.toml
jemoreira: The fact that the code uses an empty string is an implementation detail. The real outcome of this setting is that no credentials are used to communicate with the build api.
Changed the order as requested.

build/package/host/etc/cvdr.toml
jemoreira: Done

ser-io (approved):

PR #247, opened 2023-11-21
Adds `get_config` command.

ser-io:
- The new command is an utility to fetch the final configuration parameter values.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -28,6 +28,7 @@ import (
 	client "github.com/google/cloud-android-orchestration/pkg/client"
 	wclient "github.com/google/cloud-android-orchestration/pkg/webrtcclient"
 
+	"github.com/PaesslerAG/jsonpath"
 	"github.com/hashicorp/go-multierror"
 	"github.com/spf13/cobra"
 	"golang.org/x/term"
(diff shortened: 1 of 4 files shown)


pkg/cli/cli.go
```diff
@@ -1064,6 +1073,24 @@ func runDisconnectCommand(flags *ConnectFlags, c *command, args []string, opts *
 	return merr
 }
 
+func runGetConfigCommand(c *cobra.Command, args []string, cfg Config) error {
+	if len(args) == 0 {
+		return errors.New("missing parameter")
```
jemoreira, line 1078: nit: missing config property name

pkg/cli/cli.go
```diff
@@ -1064,6 +1073,24 @@ func runDisconnectCommand(flags *ConnectFlags, c *command, args []string, opts *
...
+	}
+	if len(args) > 1 {
+		return errors.New("use one parameter at a time")
```
jemoreira, line 1081: nit: same

pkg/cli/cli.go
```diff
@@ -1064,6 +1073,24 @@ func runDisconnectCommand(flags *ConnectFlags, c *command, args []string, opts *
...
+	}
+	switch args[0] {
```
jemoreira, line 1083: nit: You could implement this once and never have come back to it again by encoding the config object as JSON and then decoding it using a `map[string]interface{}`.
You'd have to split the "property name" by `'.'` to support inner fields, but that shouldn't be too hard.
We are already planning to support multiple service urls, which means the config structure could change, so don't reply with "YAGNI" :)

jemoreira (approved):

pkg/cli/cli.go

pkg/cli/cli.go

pkg/cli/cli.go
ser-io: :) done.

pkg/cli/config.go
```diff
@@ -27,22 +27,22 @@ import (
...
 
 type Config struct {
-	ServiceURL                string
-	Zone                      string
-	HTTPProxy                 string
-	ConnectionControlDir      string
-	KeepLogFilesDays          int
-	BuildAPICredentialsSource string
-	Host                      HostConfig
+	ServiceURL                string     `json:"service_url,omitempty"`
```
jemoreira, line 39: Doesn't this break parsing the config files?
The config file has `ServiceURL` instead of `service_url`

ser-io: it does not break, the toml parsing do not uses the json annotations.

PR #267, opened 2024-02-02
Handles "europe-west" and "us-central" GAE region names.

ser-io:
From: https://cloud.google.com/appengine/docs/standard/locations

```
Note: Two locations, which are called europe-west and us-central in App Engine commands and in the Google Cloud console, are called europe-west1 and us-central1, respectively, elsewhere in Google documentation.
```

### DIFF
--- a/scripts/deploy/gae/main.sh
+++ b/scripts/deploy/gae/main.sh
@@ -58,9 +58,16 @@ if [ $? -ne 0 ]; then
   gcloud app create
 fi
 
-gcloud services enable vpcaccess.googleapis.com
-
 app_region=$(gcloud app describe --format="value(locationId)")
+# Note: Two locations, which are called europe-west and us-central in App Engine commands and in the Google Cloud
+# console, are called europe-west1 and us-central1, respectively, elsewhere in Google documentation.
+# https://cloud.google.com/appengine/docs/standard/locations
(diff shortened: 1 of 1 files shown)


scripts/deploy/gae/main.sh
```diff
@@ -58,9 +58,16 @@ if [ $? -ne 0 ]; then
   gcloud app create
 fi
 
-gcloud services enable vpcaccess.googleapis.com
-
 app_region=$(gcloud app describe --format="value(locationId)")
+# Note: Two locations, which are called europe-west and us-central in App Engine commands and in the Google Cloud
+# console, are called europe-west1 and us-central1, respectively, elsewhere in Google documentation.
+# https://cloud.google.com/appengine/docs/standard/locations
+if [ "${app_region}" = "europe-west" ] || [ "${app_region}" = "us-central" ]; then
```
jemoreira, line 65: If the script is run by a human and the location is provided by them, then it's better to print a message suggesting the correct location and fail.

jemoreira (approved):

scripts/deploy/gae/main.sh
ser-io: this seems to be a glitch between App Engine and the rest of Google Cloud, where  App Engine uses name the name `us-central` and the rest of GCP expects `us-central1`. this value is never provided by a human, it will be what App Engine gives you.

PR #269, opened 2024-02-15
Create skeleton code for IM(InstanceManager) Type - docker

0405ysj:
Currently, there's no sufficient type to support containerized instance.
Of course we cannot use GCE IM type for ARM servers based on Docker.
Also using Unix IM type is restricted, since it's designed for supporting single Host Orchestrator.

In my idea, docker IM Type aims:
- Single Zone (Server itself)
- Multiple Host (One Host - One Docker instance)


0405ysj:
Since I'm newbie on this project, I cannot ensure even if my code is correct or wrong, beyond the build phase! I'm happy to get any reviews on here :)

ikicha (approved):

PR #271, opened 2024-02-20
Implement ListZones for DockerInstanceManager

0405ysj:


pkg/app/instances/docker.go
```diff
@@ -39,25 +39,44 @@ func NewDockerInstanceManager(cfg Config) *DockerInstanceManager {
...
+	return &apiv1.ListZonesResponse{
+		Items: []*apiv1.Zone{{
+			Name: "local",
+		}},
+	}, nil
 }
 
-func (m *DockerInstanceManager) CreateHost(_ string, _ *apiv1.CreateHostRequest, _ accounts.User) (*apiv1.Operation, error) {
+func (m *DockerInstanceManager) CreateHost(zone string, _ *apiv1.CreateHostRequest, _ accounts.User) (*apiv1.Operation, error) {
+	if zone != "local" {
```
jemoreira, line 50: nit: You can make this code more future proof by adding a private function that returns the list of zones ([]string). Then you use that to build the return object for `ListZones()` and in these functions to check if the zone is valid.

jemoreira (approved):

pkg/app/instances/docker.go
0405ysj: I think the best way would be modifying `getZone` and `getHost` in `pkg/app/app.go` to check the list with `instanceManager.listZones` and `instanceManager.listHosts`. I'll work on refactoring when I have time.

PR #273, opened 2024-02-21
Tool to export CF host images to GCE.

ser-io:


scripts/gcp/create_host_image.sh
```diff
@@ -0,0 +1,118 @@
...
+    \?)
+      echo "Invalid option: ${OPTARG}" >&2
+      usage
+      exit 1
+      ;;
+    :)
+      echo "Invalid option: ${OPTARG} requires an argument" >&2
+      usage
+      exit 1
+      ;;
```
jemoreira, line 63: nit: These two could be combined into
```
    \?)
      usage; exit 1
```
If you remove the starting `:` from the option string. The default messages printed by `getopts` to stderr are very similar to yours.

jemoreira (commented):
LGTM but the script is quite complex, especially the curl in a loop with json parsing. Have you considered writing this in go like other scripts we have in aosp?

scripts/gcp/create_host_image.sh
ser-io: Done.

ser-io:
> LGTM but the script is quite complex, especially the curl in a loop with json parsing. Have you considered writing this in go like other scripts we have in aosp?

yes, the pagination logic due the lack of a search endpoint in the Github Artifacts API does not look appealing in bash. However the script story is simple: "search token" -> "download" -> "upload" which I still think is a good fit for bash.

jemoreira (approved):

PR #274, opened 2024-02-21
Let the script create and cleanup the upload bucket.

ser-io:


scripts/gcp/create_host_image.sh
```diff
@@ -102,10 +100,11 @@ curl -L \
 
 unzip image.zip
 
-gcloud storage cp image.tar.gz  gs://${gce_bucket}/${name}.tar.gz
+gcloud storage buckets create gs://${bucket_name} --location="us-east1" --project="${project_id}"
```
jemoreira, line 103: location should be configurable

jemoreira (approved):

scripts/gcp/create_host_image.sh
ser-io: ack.

PR #278, opened 2024-03-08
Sanitize auth token value.

ser-io:
- Remove end of line ('\n') characters, if exists, in oidc token values.


pkg/cli/cli.go
```diff
@@ -1182,7 +1182,7 @@ func buildServiceBuilder(builder client.ServiceBuilder, authnConfig *AuthnConfig
 			}
 			opts.Authn = &client.AuthnOpts{
 				OIDCToken: &client.OIDCToken{
-					Value: string(token),
+					Value: strings.TrimSuffix(string(token), "\n"),
```
jemoreira, line 1185: nit: This should be done right after reading it from the file, so that the token variable holds a valid token all the time.

jemoreira (approved):

PR #286, opened 2024-03-20
Implement WaitOperation of dockerIMType

0405ysj:
This PR is needed to support using `cvdr host create` and `cvdr host delete`.

### DIFF
--- a/pkg/app/instances/docker.go
+++ b/pkg/app/instances/docker.go
@@ -43,6 +43,9 @@ type DockerInstanceManager struct {
 	Client client.Client
 }
 
+const CreateHostOPType string = "createhost"
+const DeleteHostOPType string = "deletehost"
+
 func NewDockerInstanceManager(cfg Config, cli client.Client) *DockerInstanceManager {
 	return &DockerInstanceManager{
 		Config: cfg,
@@ -81,7 +84,7 @@ func (m *DockerInstanceManager) CreateHost(zone string, _ *apiv1.CreateHostReque
 		return nil, fmt.Errorf("Failed to start docker container: %w", err)
 	}
 	return &apiv1.Operation{
-		Name: "createhost_" + createRes.ID,
+		Name: EncodeOperationName(CreateHostOPType, createRes.ID),
 		Done: true,
 	}, nil
 }
@@ -120,24 +123,43 @@ func (m *DockerInstanceManager) DeleteHost(zone string, _ accounts.User, host st
 		return nil, fmt.Errorf("Failed to remove docker container: %w", err)
 	}
 	return &apiv1.Operation{
-		Name: "deletehost_" + host,
+		Name: EncodeOperationName(DeleteHostOPType, host),
 		Done: true,
 	}, nil
 }
 
+func EncodeOperationName(opType string, host string) string {
+	return opType + "_" + host
+}
+
+func DecodeOperationName(name string) (string, string, error) {
+	words := strings.SplitN(name, "_", 2)
+	if len(words) == 2 {
+		return words[0], words[1], nil
+	} else {
+		return "", "", errors.NewBadRequestError(fmt.Sprintf("cannot parse operation from %q", name), nil)
+	}
+}
+
 func (m *DockerInstanceManager) waitCreateHostOperation(host string) (*apiv1.HostInstance, error) {
-	ctx := context.TODO()
+	ctx, cancel := context.WithTimeout(context.TODO(), 3*time.Minute)
+	defer cancel()
 	for {
-		res, err := m.Client.ContainerInspect(ctx, host)
-		if err != nil {
-			return nil, fmt.Errorf("Failed to inspect docker container: %w", err)
+		select {
+		case <-ctx.Done():
+			return nil, errors.NewServiceUnavailableError("Wait for operation timed out", nil)
+		default:
+			res, err := m.Client.ContainerInspect(ctx, host)
+			if err != nil {
+				return nil, fmt.Errorf("Failed to inspect docker container: %w", err)
+			}
+			if res.State.Running {
+				return &apiv1.HostInstance{
+					Name: host,
+				}, nil
+			}
+			time.Sleep(time.Second)
 		}
-		if res.State.Running {
-			return &apiv1.HostInstance{
-				Name: host,
-			}, nil
-		}
-		time.Sleep(time.Second)
 	}
 }
 
@@ -161,14 +183,17 @@ func (m *DockerInstanceManager) WaitOperation(zone string, _ accounts.User, name
 	if zone != "local" {
 		return nil, errors.NewBadRequestError("Invalid zone. It should be 'local'.", nil)
 	}
-	words := strings.SplitN(name, "_", 2)
-	switch words[0] {
-	case "createhost":
-		return m.waitCreateHostOperation(words[1])
-	case "deletehost":
-		return m.waitDeleteHostOperation(words[1])
+	opType, host, err := DecodeOperationName(name)
+	if err != nil {
+		return nil, err
+	}
+	switch opType {
+	case CreateHostOPType:
+		return m.waitCreateHostOperation(host)
+	case DeleteHostOPType:
+		return m.waitDeleteHostOperation(host)
 	default:
-		return nil, errors.NewBadRequestError(fmt.Sprintf("operation %q not found.", name), nil)
+		return nil, errors.NewBadRequestError(fmt.Sprintf("operation type %s not found.", opType), nil)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


ikicha:
It looks like every API(at least APIs which we use) is synchronous, so I think we don't need to wait in "wait", every job is done when the method returns result.

0405ysj:
I think implementing `WaitOperation` like current is required to satisfy the definition of API & support `cvdr host` commands. However, I agree with you to set response of `createhost` or `deletehost` as `Done:true` if they're all synchronous.

I'll check tomorrow as well if they're synchronous, and change `Done` value from false to true if it is.

pkg/app/instances/docker.go
```diff
@@ -78,8 +80,8 @@ func (m *DockerInstanceManager) CreateHost(zone string, _ *apiv1.CreateHostReque
 		return nil, fmt.Errorf("Failed to start docker container: %w", err)
 	}
 	return &apiv1.Operation{
-		Name: createRes.ID,
-		Done: true,
+		Name: "createhost_" + createRes.ID,
```
ser-io, line 83: let's create constants for `createhost` and `deletehost`.

pkg/app/instances/docker.go
```diff
@@ -117,16 +119,56 @@ func (m *DockerInstanceManager) DeleteHost(zone string, _ accounts.User, host st
...
+		}, nil
+	}
+}
+
+func (m *DockerInstanceManager) WaitOperation(zone string, _ accounts.User, name string) (any, error) {
 	if zone != "local" {
 		return nil, fmt.Errorf("Invalid zone. It should be 'local'.")
 	}
-	return nil, fmt.Errorf("%T#WaitOperation is not implemented", *m)
+	words := strings.SplitN(name, "_", 2)
```
ser-io, line 163: let's isolate the logic of encoding operation names and decoding operation names and add unit test to it.

we could have two functions: `EncodeOperationName(type, host string)` and `DecodeOperationName(name string) (string, string, error)`. Feel free to create a new package `docker` as these function names might be common in future instance manager implementations

this would help us be more robust and avoid crashing down below when using `words[1]` if `words` has length 1 for example.

pkg/app/instances/docker.go
```diff
@@ -117,16 +119,56 @@ func (m *DockerInstanceManager) DeleteHost(zone string, _ accounts.User, host st
...
 	return &apiv1.Operation{
-		Name: host,
-		Done: true,
+		Name: "deletehost_" + host,
+		Done: false,
 	}, nil
 }
 
-func (m *DockerInstanceManager) WaitOperation(zone string, _ accounts.User, _ string) (any, error) {
+func (m *DockerInstanceManager) WaitCreateHostOperation(host string) (any, error) {
```
ser-io, line 127: make it private `waitCreateHostOperation`.

pkg/app/instances/docker.go
```diff
@@ -117,16 +119,56 @@ func (m *DockerInstanceManager) DeleteHost(zone string, _ accounts.User, host st
...
+		if res.State.Running {
+			return &apiv1.HostInstance{
+				Name: host,
+			}, nil
+		}
+		time.Sleep(time.Second)
+	}
+}
+
+func (m *DockerInstanceManager) WaitDeleteHostOperation(host string) (any, error) {
```
ser-io, line 143: make it private `waitDeleteHostOperation`

pkg/app/instances/docker.go
```diff
@@ -117,16 +119,56 @@ func (m *DockerInstanceManager) DeleteHost(zone string, _ accounts.User, host st
...
+	ctx, cancel := context.WithTimeout(context.TODO(), 3*time.Minute)
+	defer cancel()
+	resCh, errCh := m.Client.ContainerWait(ctx, host, "")
+	select {
+	case <-ctx.Done():
+		return nil, fmt.Errorf("Timeout while waiting for deleting host")
```
ser-io, line 149: nit: let's use `errors.New()` instead as we are not using format verbs.

pkg/app/instances/docker.go
ser-io, line 127: we can return `*apiv1.HostInstance` here, same in `WaitDeleteHostOperation`

pkg/app/instances/docker.go
ser-io, line 149: use our `errors` package `github.com/google/cloud-android-orchestration/pkg/app/errors` i.e`  `return nil, errors.NewServiceUnavailableError("Wait for operation timed out", nil)`

pkg/app/instances/docker.go
```diff
@@ -117,16 +119,56 @@ func (m *DockerInstanceManager) DeleteHost(zone string, _ accounts.User, host st
...
 	}
-	return nil, fmt.Errorf("%T#WaitOperation is not implemented", *m)
+	words := strings.SplitN(name, "_", 2)
+	switch words[0] {
+	case "createhost":
+		return m.WaitCreateHostOperation(words[1])
+	case "deletehost":
+		return m.WaitDeleteHostOperation(words[1])
+	default:
+		return nil, fmt.Errorf("%T#WaitOperation for %s is not implemented", *m, words[0])
```
ser-io, line 170: let's use the app errors, i.e: `errors.NewBadRequestError(fmt.Sprintf("operation %q not found.", name), nil)`

pkg/app/instances/docker.go
```diff
@@ -117,16 +119,56 @@ func (m *DockerInstanceManager) DeleteHost(zone string, _ accounts.User, host st
...
-		Done: true,
+		Name: "deletehost_" + host,
+		Done: false,
 	}, nil
 }
 
-func (m *DockerInstanceManager) WaitOperation(zone string, _ accounts.User, _ string) (any, error) {
+func (m *DockerInstanceManager) WaitCreateHostOperation(host string) (any, error) {
+	ctx := context.TODO()
+	for {
```
ser-io, line 129: let's not have an infinite loop here. let's have a fixed number of retries, if the loop ends we return `create failed`.

ser-io (changes requested):

pkg/app/instances/docker.go
0405ysj: Used errors in `pkg/app/errors.go`, since it seemed it's standard one in our github repo.

pkg/app/instances/docker.go
0405ysj: Instead introducing fixed number of retries, set timeout for the function.

pkg/app/instances/docker.go
0405ysj: Introduced `EncodeOperationName` and `DecodeOperationName`, and added unit tests for them.

pkg/app/instances/docker.go
```diff
@@ -117,16 +123,78 @@ func (m *DockerInstanceManager) DeleteHost(zone string, _ accounts.User, host st
...
+	if len(words) == 2 {
+		return words[0], words[1], nil
+	} else {
+		return "", "", errors.NewBadRequestError(fmt.Sprintf("cannot parse operation from %q", name), nil)
+	}
+}
+
+func (m *DockerInstanceManager) waitCreateHostOperation(host string) (*apiv1.HostInstance, error) {
+	ctx, cancel := context.WithTimeout(context.TODO(), 3*time.Minute)
+	defer cancel()
```
ikicha, line 146: If every API we use is sync, we don't need timeout, just check and then return the result.

pkg/app/instances/docker.go
ser-io: better, thanks!

pkg/app/instances/docker.go
```diff
@@ -43,6 +43,9 @@ type DockerInstanceManager struct {
 	Client client.Client
 }
 
+const CreateHostOPType string = "createhost"
```
ser-io, line 46: let's use a type alias here and a `const` block

```
type OpType string

const (
	CreateHostOpType OpType = "createhost"
	DeleteHostOPType OpType = "deletehost"
)
```

ser-io (approved):

pkg/app/instances/docker.go
0405ysj: In our usages, I think your opinion is right. However because of definition of this API, I think it's still better to have timeout mechanism if it doesn't harm any correctness or performance issue.

But I prefer to follow your opinion if you still want, since it has no actual usage. HDYT? @ikicha

pkg/app/instances/docker.go
0405ysj: Defined `OPType`.

pkg/app/instances/docker.go
ikicha: I am just worried that it needs to wait until timeout for an error case even though there is no chance to be recovered

PR #291, opened 2024-03-22
Fallback to upload current dir when running cvdr create with local_image

0405ysj:
Testing method for the ARM server:
```
cvdr \
--http_proxy=socks5://localhost:1337 \
--service_url=http://100.114.242.57:8080 \
--zone=local \
--host=$CO_HOST_NAME \
--local_image=true \
--auto_connect=false \
create
```

### DIFF
--- a/pkg/cli/cvd.go
+++ b/pkg/cli/cvd.go
@@ -143,17 +143,30 @@ func (c *cvdCreator) Create() ([]*hoapi.CVD, error) {
 
 func (c *cvdCreator) createCVDFromLocalBuild() ([]*hoapi.CVD, error) {
 	vars, err := GetAndroidEnvVarValues()
-	if err != nil {
-		return nil, fmt.Errorf("error retrieving Android Build environment variables: %w", err)
+	var names []string
+	if err == nil {
+		images, err := ListLocalImageRequiredFiles(vars)
+		if err != nil {
+			return nil, fmt.Errorf("error building list of required image files: %w", err)
+		}
+		names = images
+		if err := verifyCVDHostPackageTar(vars.HostOut); err != nil {
+			return nil, fmt.Errorf("invalid cvd host package: %w", err)
+		}
+		names = append(names, filepath.Join(vars.HostOut, CVDHostPackageName))
+	} else {
+		currentDir, err := os.Getwd()
+		if err != nil {
+			return nil, fmt.Errorf("failed to get current directory: %w", err)
+		}
+		files, err := os.ReadDir(currentDir)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read current directory: %w", err)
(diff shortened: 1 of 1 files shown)


0405ysj:
Interested thing was, it took only about 5-7 mins to upload artifacts & running CF instance. Probably, would it be caused by using public stun server(stun.l.google.com)?

ikicha:
Interesting, I think it might be due to socks5 proxy rather than stun server. can you benchmark network bandwidth with socks5?

0405ysj:
I roughly checked with our hybrid one, and it took about 12-13 mins. And `super.img` was about 4.7G. Seems like the bandwidth is about 6MB/s or 48Mbps

pkg/cli/cvd.go
```diff
@@ -143,17 +143,30 @@ func (c *cvdCreator) Create() ([]*hoapi.CVD, error) {
...
-		return nil, fmt.Errorf("error retrieving Android Build environment variables: %w", err)
-	}
-	names, err := ListLocalImageRequiredFiles(vars)
-	if err != nil {
-		return nil, fmt.Errorf("error building list of required image files: %w", err)
-	}
-	if err := verifyCVDHostPackageTar(vars.HostOut); err != nil {
-		return nil, fmt.Errorf("invalid cvd host package: %w", err)
+	var names []string
+	if err == nil {
```
jemoreira, line 147: The code pattern for this function should be roughly:
```
cvd, err := createCVDFromBuiltSource(...)
if errors.Is(err, NotInAndroidSourceErr) {
  return createCVDFromCurrentDir(...)
}
return cvd, err
```
I think encapsulating both flows into their own functions will increase readability.

jemoreira (changes requested):

ser-io:
Uploading everything from the current directory with no validation is a big responsibility to pass to the final user, this could end up in undesirable behaviors, for example uploading personal files.

ikicha:
IIUC, imgs.zip has required_image file inside. We can use that to list up files which we want to send.

0405ysj:
> Wrote https://github.com/google/cloud-android-orchestration/pull/292 where we need to be upfront with what's going to be uploaded, it also offers more flexibility to build instances with mixed artifacts, as we just need to change the artifacts sources file.

Should I close this PR then? I saw the PR you've created, and this and that contains duplicated contents. Seems your PR covers wider purposes.


> IIUC, imgs.zip has required_image file inside. 

Do we? I unpacked CF image zip file downloaded from ci.android.com, and there was no `required_image`.

ikicha:
ah, sorry, require_image exists in PRODUCT_OUT, not imgs.zip

ser-io:
yes, let's close this one, thanks!

PR #292, opened 2024-03-22
Create instances with local artifacts.

ser-io:


pkg/cli/cli.go
```diff
@@ -491,15 +492,20 @@ func cvdCommands(opts *subCommandOpts) []*cobra.Command {
 	// Local image
 	create.Flags().BoolVar(&createFlags.LocalImage, localImageFlag, false,
 		"Builds a CVD with image files built locally, the required files are https://cs.android.com/android/platform/superproject/+/master:device/google/cuttlefish/required_images and cvd-host-packages.tar.gz")
-	localImgMutuallyExFlags := []string{
+	// Local artifacts flag
+	create.Flags().StringVar(&createFlags.LocalArtifacts, localArtifacsFlag, "",
+		"Path to file containing the local artifacts sources")
```
0405ysj, line 497: I think it's better to have more detailed information about what `local artifacts sources` means, such as `toml file containing paths of local artifacts`.

pkg/cli/cvd.go
```diff
@@ -412,3 +450,67 @@ func verifyCVDHostPackageTar(dir string) error {
...
+	return result
+}
+
+func loadLocalArtifacsSrcs(path string) (*LocalArtifactsSrcs, error) {
+	b, err := os.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	decoder := toml.NewDecoder(bytes.NewReader(b))
+	r := &LocalArtifactsSrcs{}
```
0405ysj, line 502: IIUC, toml contains metadata of CF images & cvd-host_package.tar.gz. Could we have example toml file so that users can make files by themselves?

pkg/cli/cvd.go
```diff
@@ -412,3 +450,67 @@ func verifyCVDHostPackageTar(dir string) error {
...
+		result = append(result, v)
+	}
+	return result
+}
+
+func loadLocalArtifacsSrcs(path string) (*LocalArtifactsSrcs, error) {
+	b, err := os.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
```
0405ysj, line 500: What about applying default configuration if user doesn't have toml file, but want to `cvdr create` with `local_artifacts`?

pkg/cli/cvd.go
```diff
@@ -412,3 +450,67 @@ func verifyCVDHostPackageTar(dir string) error {
...
+func (s *LocalArtifactsSrcs) validate() error {
+	if s.Bootloader == "" {
+		return errors.New("missing bootloader source")
+	}
+	if s.CVDHostPackageTar == "" {
+		return errors.New("missing cvd host package tar source")
+	}
+
+	// TODO: Make it configurable.
+	var allowedImageKeys = []string{
```
0405ysj, line 469: How about using `map[string]bool`, not to introduce `contains` function?

pkg/cli/cvd.go
```diff
@@ -412,3 +450,67 @@ func verifyCVDHostPackageTar(dir string) error {
...
 	return nil
 }
+
+type LocalArtifactsSrcs struct {
+	Images            map[string]string
+	Bootloader        string
+	CVDHostPackageTar string
+}
+
+func (s *LocalArtifactsSrcs) validate() error {
```
0405ysj, line 460: OOC, would we need to validate if essential CF images exist or not? For example, checking whether `vbmeta.img` is defined in `LocalArtifactsSrcs` or not.

pkg/cli/cvd.go
```diff
@@ -247,6 +252,39 @@ func (c *cvdCreator) createWithOpts() ([]*hoapi.CVD, error) {
 	return res.CVDs, nil
 }
 
+func (c *cvdCreator) createCVDFromLocalArtifacts() ([]*hoapi.CVD, error) {
+	fmt.Println("here createCVDFromLocalArtifacts")
```
0405ysj, line 256: Remove this?

pkg/cli/cvd.go
```diff
@@ -247,6 +252,39 @@ func (c *cvdCreator) createWithOpts() ([]*hoapi.CVD, error) {
...
+	srcs, err := loadLocalArtifacsSrcs(c.opts.LocalArtifacts)
+	if err != nil {
+		return nil, fmt.Errorf("failed loading local artifacts file: %w", err)
+	}
+	if err := srcs.validate(); err != nil {
+		return nil, fmt.Errorf("invalid local artifacts: %w", err)
+	}
+	uploadDir, err := c.service.HostService(c.opts.Host).CreateUploadDir()
```
0405ysj, line 264: From this part, I think all logics are same as `createCVDFromLocalBuild`. Could we unite logic with introducing a new function?

0405ysj (changes requested):

pkg/cli/cvd.go
ser-io: true, however let's have the refactoring in a future PR as I'm working on changes on the upload API to handle compression before uploads.

pkg/cli/cvd.go
ser-io: I see, go 1.21 introduced the [`slices` package](https://pkg.go.dev/slices) which have a `Contain` implementation, look forward to upgrade to repo to go 1.21 soon.

pkg/cli/cvd.go
ser-io: hmm, it's not clear to me what a default config would look like here. what's your suggestion here?

pkg/cli/cvd.go
ser-io: good point, rephrased the flag documentation adding the example there.

pkg/cli/cvd.go
ser-io: good point, however I couldn't find any external documentation listing the required partition image for a cuttlefish device to give the user more context here, let's keep it flexible for now, we can always add this sort of validation later on.

pkg/cli/cli.go
```diff
@@ -87,6 +87,7 @@ const (
 	buildIDFlag               = "build_id"
 	buildTargetFlag           = "build_target"
 	localImageFlag            = "local_image"
+	localArtifacsFlag         = "local_artifacts"
```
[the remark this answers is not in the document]

pkg/cli/cvd.go
ser-io: good catch, done.

pkg/cli/cvd.go
0405ysj: Okay, I also agree with you. We don't need this kind of validation immediately. If there's no rule on here, then it's fine for me :)

pkg/cli/cvd.go
0405ysj: Conceptually, assuming someone set `--local_artifacts=default`, then `loadLocalArtifactsSrcs` parses the flag, recognizes `default`, and creates a default `LocalArtifactsSrcs` struct object not reading a file from given path.

On the other hand, I'm not certain whether setting default cfg is good idea or not, in case of aosp_cf someday requires another image(e.g. vendor.img) which default config doesn't know.

pkg/cli/cvd.go
0405ysj: :+1:

pkg/cli/cvd.go
0405ysj: Ah, your idea sounds better, if golang 1.21 supports that. Could you just leave a TODO comment in this PR then?

pkg/cli/cvd.go
0405ysj: Okay. Sounds good. Just let this comment thread be opened. Leaving a TODO comment might be better, to grab refactoring chance by someone in the future.

0405ysj (approved):
Though having minor comments, I don't have any strong opinions leading large fixes.

pkg/cli/cvd.go
```diff
@@ -247,6 +252,38 @@ func (c *cvdCreator) createWithOpts() ([]*hoapi.CVD, error) {
 	return res.CVDs, nil
 }
 
+func (c *cvdCreator) createCVDFromLocalArtifacts() ([]*hoapi.CVD, error) {
+	srcs, err := loadLocalArtifactsSrcs(c.opts.LocalArtifacts)
```
ikicha, line 256: instead of list up necessary artifacts in toml file, how about using *-imgs.zip as HO for AB and fetch_cvd does?

ser-io: `ImagesZip` is a valid option as an entry for the local artifacts config file as it's what AB gives you, I'm going to add it in a follow up PR. In such case the config file would have two entries only:

```
ImagesZip="PATH/-imgs.zip"
CVDHostPackageTar="PATH/cvd-host_package.tar.gz"
```

pkg/cli/cli.go
```diff
@@ -491,15 +492,31 @@ func cvdCommands(opts *subCommandOpts) []*cobra.Command {
...
+CVDHostPackageTar="PATH/cvd-host_package.tar.gz"
+
+[Images]
+boot = "PATH/boot.img"
+init_boot = "PATH/init_boot.img"
+super = "PATH/super.img"
+userdata = "PATH/userdata.img"
+` + "```"
+	create.Flags().StringVar(&createFlags.LocalArtifacts, localArtifactsFlag, "", d)
+	localBuildMutuallyExFlags := []string{
```
jemoreira, line 509: nit: This name is a mouthful and doesn't make the code any clearer. Consider having two lists: one for remote build flags and one for local build flags; and then two nested loops to mark them mutually exclusive.

pkg/cli/cli.go
```diff
@@ -491,15 +492,31 @@ func cvdCommands(opts *subCommandOpts) []*cobra.Command {
...
 		branchFlag, buildIDFlag, buildTargetFlag,
 		kernelBranchFlag, kernelBuildIDFlag, kernelBuildTargetFlag,
 		bootloaderBranchFlag, bootloaderBuildIDFlag, bootloaderBuildTargetFlag,
 		systemImgBranchFlag, systemImgBuildIDFlag, systemImgBuildTargetFlag,
 	}
-	for _, f := range localImgMutuallyExFlags {
+	for _, f := range localBuildMutuallyExFlags {
 		create.MarkFlagsMutuallyExclusive(f, localImageFlag)
+		create.MarkFlagsMutuallyExclusive(f, localArtifactsFlag)
```
jemoreira, line 517: Shouldn't these two be mutually exclusive too?

The only reason I can image these not being mutually exclusive is using a combination of locally built host tools with downloaded/built-elsewhere guest, but that same argument applies to locally built host with guest from build api, so nothing would be mutually exclusive then.

pkg/cli/cli.go
```diff
@@ -87,6 +87,7 @@ const (
 	buildIDFlag               = "build_id"
 	buildTargetFlag           = "build_target"
 	localImageFlag            = "local_image"
+	localArtifactsFlag        = "local_artifacts"
```
jemoreira, line 90: nit: Add _file to the flag to make it clear that it points to a file describing the artifacts, not the artifacts themselves.

jemoreira (commented):
I think this approach is not very user friendly. Having to write a file just to be to upload artifacts seems like too much work, especially if the file is not that much useful afterwards.
What's the reason to use an artifacts file instead of a set of flags (one per property supported in the file)?

pkg/cli/cvd.go
ser-io: yes a TODO comment was needed, however we are no longer doing this checking after changing the sources values to flags.

pkg/cli/cli.go
ser-io: good point, done

pkg/cli/cli.go
ser-io: yes, these flags  are mutually exclusive, we had line `create.MarkFlagsMutuallyExclusive(localImageFlag, localArtifactsFlag)` down below, the code changed after using flags, however the mutual exclusion part stays.

pkg/cli/cli.go
ser-io: no longer needed after changing to flags.

ser-io:
@jemoreira good point on using flags rather a file, I made the relevant changes. PTAL. thanks!

jemoreira (approved):

PR #295, opened 2024-03-28
Change HostOut path for ARM64 during cvdr create

0405ysj:
After running `lunch`, ANDROID_HOST_OUT is set as `$ANDROID_BUILD_TOP/out/host/linux-x86`. To be run in ARM64 dockerized instance, we should send `cvd-host_package.tar.gz` for ARM64.

### DIFF
--- a/pkg/cli/cvd.go
+++ b/pkg/cli/cvd.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 
@@ -163,10 +164,15 @@ func (c *cvdCreator) createCVDFromLocalBuild() ([]*hoapi.CVD, error) {
 	if err != nil {
 		return nil, err
 	}
-	hostOut, err := envVar(AndroidHostOutVarName)
+	targetArch, err := getTargetArch(buildTop)
 	if err != nil {
 		return nil, err
 	}
+	hostOutRelativePath, err := GetHostOutRelativePath(targetArch)
+	if err != nil {
+		return nil, err
+	}
+	hostOut := filepath.Join(buildTop, hostOutRelativePath)
 	if err := verifyCVDHostPackageTar(hostOut); err != nil {
 		return nil, err
 	}
@@ -395,7 +401,6 @@ const (
 
 const (
 	AndroidBuildTopVarName   = "ANDROID_BUILD_TOP"
-	AndroidHostOutVarName    = "ANDROID_HOST_OUT"
 	AndroidProductOutVarName = "ANDROID_PRODUCT_OUT"
 )
 
@@ -419,6 +424,29 @@ func ListLocalImageRequiredFiles(buildTop, productOut string) ([]string, error) 
 		result = append(result, filepath.Join(productOut, line))
 	}
 	return result, nil
+}
+
+func getTargetArch(buildTop string) (string, error) {
+	// `$ANDROID_BUILD_TOP/out/soong_ui` can bring values of build variables, set by `lunch` command.
+	// https://cs.android.com/android/platform/superproject/main/+/main:build/soong/cmd/soong_ui/main.go;l=298
+	bin := filepath.Join(buildTop, "out/soong_ui")
+	cmdOut, err := exec.Command(bin, "--dumpvar-mode", "TARGET_ARCH").Output()
+	if err != nil {
+		return "", fmt.Errorf("error while getting target arch: %w", err)
+	}
+	return strings.TrimSpace(string(cmdOut)), nil
+}
+
+func GetHostOutRelativePath(targetArch string) (string, error) {
+	m := map[string]string{
+		"x86_64": "out/host/linux-x86",
+		"arm64":  "out/host/linux_musl-arm64",
+	}
+	relativePath, ok := m[targetArch]
+	if !ok {
+		return "", fmt.Errorf("Unexpected target architecture: %q", targetArch)
+	}
+	return relativePath, nil
 }
 
 func verifyCVDHostPackageTar(dir string) error {
--- a/pkg/cli/cvd_test.go
+++ b/pkg/cli/cvd_test.go
@@ -71,3 +71,27 @@ func TestAdditionalInstancesNum(t *testing.T) {
 		}
 	}
 }
+
+func TestGetHostOutRelativePathSucceeds(t *testing.T) {
+	targetArch := "x86_64"
+
+	got, err := GetHostOutRelativePath(targetArch)
+
+	if err != nil {
+		t.Fatal(err)
+	}
+	expected := "out/host/linux-x86"
+	if diff := cmp.Diff(expected, got); diff != "" {
+		t.Errorf("mismatch (-want +got):\n%s", diff)
+	}
+}
+
+func TestGetHostOutRelativePathFailsUnknownTargetArch(t *testing.T) {
+	targetArch := "unknown"
+
+	_, err := GetHostOutRelativePath(targetArch)
+
+	if err == nil {
+		t.Errorf("expected error")
+	}
+}


0405ysj:
WIP for syncing with existing test.

0405ysj:
Done for WIP, now resolved test failure in presubmit test.

pkg/cli/cvd.go
```diff
@@ -155,6 +156,9 @@ func (c *cvdCreator) createCVDFromLocalBuild() ([]*hoapi.CVD, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error retrieving Android Build environment variables: %w", err)
 	}
+	if err = vars.regenerateHostOut(); err != nil {
```
ser-io, line 159: the logic within method `regenerateHostOut` is about getting the target architecture and then override its `HostOut` field. Let's isolate the functionality of getting the target architecture and move it to the end of the file and modify `GetAndroidEnvVarValues`  to have it returning the relevant HostOut value.

ser-io (changes requested):

pkg/cli/cvd.go
```diff
@@ -415,6 +427,20 @@ func GetAndroidEnvVarValues() (AndroidEnvVars, error) {
 	}, nil
 }
 
+func (v *AndroidEnvVars) regenerateHostOut() error {
+	cmdOut, err := exec.Command(v.BuildTop+"/"+SoongUiBinPath, DumpvarModeFlag, AndroidTargetArchVarName).Output()
```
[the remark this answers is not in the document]
0405ysj: AFAIK, `get_build_var` is not an independent binary file, living as a function of the shell script. i tried to run with `exec.Command`, but it couldn't recognize `get_build_var`.

pkg/cli/cvd.go
0405ysj: Just for double-checking before starting on edit, your suggestion is `GetAndroidEnvVarValues` get arch information as parameter so that we could continue to maintain our unit tests, w/o relying on any binary files like `soong_ui`, right? I'll try to work on that.

0405ysj: Ah, I thought wrongly. To run `soong_ui`, it needs `ANDROID_BUILD_TOP` env. However it's extracted in `GetAndroidEnvVarValues`. So functionally,  running `soong_ui` should exist inside `GetAndroidEnvVarValues`, how should we do unit tests then? @ser-io

ser-io: I see, removed the `AndroidEnvVars` and `GetAndroidEnvVarValues` struct in https://github.com/google/cloud-android-orchestration/pull/297 to compute the values independently. After rebase, let's then create two functions: `targetArch(buildtop string)`  and `hostOut(arch string)`

0405ysj: Done again, after rebasing. I think this comment is now outdated, closing this comment.

pkg/cli/cvd.go
```diff
@@ -421,6 +434,25 @@ func ListLocalImageRequiredFiles(buildTop, productOut string) ([]string, error)
 	return result, nil
 }
 
+func getTargetArch(buildTop string) (string, error) {
+	cmdOut, err := exec.Command(buildTop+"/"+SoongUiBinPath, DumpvarModeFlag, AndroidTargetArchVarName).Output()
```
ser-io, line 438: nit: Use `filepath.Join(buildTop, "out/soong_ui")`.

pkg/cli/cvd.go
```diff
@@ -395,10 +400,18 @@ const (
 
 const (
 	AndroidBuildTopVarName   = "ANDROID_BUILD_TOP"
-	AndroidHostOutVarName    = "ANDROID_HOST_OUT"
 	AndroidProductOutVarName = "ANDROID_PRODUCT_OUT"
 )
 
+const (
```
ser-io, line 406: As these const variables are only used in one place I wouldn't create them and rather use the literals directly.

```
bin := filepath.Join(buildTop, "out/soong_ui")
cmdOut, err := exec.Command(bin, "--dumpvar-mode", "TARGET_ARCH").Output()
```

it makes it easier to read the command we are executing, otherwise we need to go back and forth. If you decide to keep the const variables, please declare them within the function making it closer to where it's used.

pkg/cli/cvd.go
```diff
@@ -395,10 +400,18 @@ const (
...
+	SoongUiBinPath           = "out/soong_ui"
+	DumpvarModeFlag          = "--dumpvar-mode"
+	AndroidTargetArchVarName = "TARGET_ARCH"
+)
+
+const AndroidHostOutX8664Path = "out/host/linux-x86"
```
ser-io, line 412: nit: use a `const` block

```
const (
    AndroidHostOutX8664Path = "out/host/linux-x86"
    AndroidHostOutArm64Path = "out/host/linux_musl-arm64"
)
```

pkg/cli/cvd.go
```diff
@@ -421,6 +434,25 @@ func ListLocalImageRequiredFiles(buildTop, productOut string) ([]string, error)
...
 
+func getTargetArch(buildTop string) (string, error) {
+	cmdOut, err := exec.Command(buildTop+"/"+SoongUiBinPath, DumpvarModeFlag, AndroidTargetArchVarName).Output()
+	if err != nil {
+		return "", fmt.Errorf("error while getting target arch: %w", err)
+	}
+	return strings.TrimSpace(string(cmdOut)), nil
+}
+
+func GetHostOut(buildTop, targetArch string) (string, error) {
```
ser-io, line 445: don't pass `buildTop` and let the caller do the concatenation.

pkg/cli/cvd.go
```diff
@@ -421,6 +434,25 @@ func ListLocalImageRequiredFiles(buildTop, productOut string) ([]string, error)
 	return result, nil
 }
 
+func getTargetArch(buildTop string) (string, error) {
```
ser-io, line 437: let's add documentation to this method, let's describe what `out/soong_ui`, how is it build and how it give us the target architecture?

What happen if `out/soong_ui` does not exist, imagine it gets deprecated, should we default to `x86_64`?

ser-io (changes requested):

pkg/cli/cvd.go
0405ysj: > let's add documentation to this method, let's describe what out/soong_ui, how is it build and how it give us the target architecture?

I think sharing this [link](https://cs.android.com/android/platform/superproject/main/+/main:build/soong/cmd/soong_ui/main.go;l=298) would be enough. And `out/soong_ui` extracts build variables exist in `.mk` file based on `lunch` setting.

> What happen if `out/soong_ui` does not exist, imagine it gets deprecated, should we default to x86_64?

Umm.. I don't think we should set `x86_64` as default. Instead, we can just throw an error like current.
Though we can just use `ANDROID_HOST_OUT` as fallback one, but I think it would make harder on debugging when we face an issue.

0405ysj: > I think sharing this [link](https://cs.android.com/android/platform/superproject/main/+/main:build/soong/cmd/soong_ui/main.go;l=298) would be enough. And out/soong_ui extracts build variables exist in .mk file based on lunch setting.

Added a link for `out/soong_ui`.

ser-io: I understand, thanks!

ser-io (changes requested):
let's squash this specific commit (Apply incoming reviews) with the first one. Feel free to use `git commit --amend` && `git push -f` when having changes addressing review comments only.

pkg/cli/cvd_test.go
```diff
@@ -72,41 +72,38 @@ func TestAdditionalInstancesNum(t *testing.T) {
...
 
 	if err != nil {
 		t.Fatal(err)
 	}
-	expected := "foo/out/host/linux_musl-arm64"
+	expected := "out/host/linux_musl-arm64"
 	if diff := cmp.Diff(expected, got); diff != "" {
 		t.Errorf("mismatch (-want +got):\n%s", diff)
 	}
 }
```
ser-io, line 101: Golang encourages the use of table driven tests https://go.dev/wiki/TableDrivenTests#introduction, which is a good fit in this case.

However, I think verifying all values within the internal lookup table is the same as writing the table itself. Due the simplicity of the function under test, I think there's no need for unit test. However if adding unit tests I think we should test two behaviors only, when the target arch is supported and when it's not rather than verifying all the the arch values.

ser-io (changes requested):

pkg/cli/cvd_test.go
0405ysj: Done, but just left 2 unit test; one for known arch and the other for unknown arch.

0405ysj:
@ser-io Since there was a conversation about squashing, what's the preferred branch strategy for this and `google/android-cuttlefish`? IIUC, for a single PR, squashing all commits just for applying review, and choosing `rebase and merge` when merge PR?

ser-io (approved):

ikicha (approved):

ser-io:
> @ser-io Since there was a conversation about squashing, what's the preferred branch strategy for this and `google/android-cuttlefish`? IIUC, for a single PR, squashing all commits just for applying review, and choosing `rebase and merge` when merge PR?

we can have multiple commits in one PR, but we should avoid commits whose message is "resolving or addressing review comments" only. yes `rebase and merge` would keep the upstream main history cleaner.

PR #298, opened 2024-03-29
Use the new `/:extract` endpoint when uploading compressed files.

ser-io:


pkg/cli/cli_test.go
```diff
@@ -118,45 +118,37 @@ func (fakeService) HostService(host string) client.HostOrchestratorService {
 
 type fakeHostService struct{}
 
-func (fakeHostService) GetInfraConfig() (*apiv1.InfraConfig, error) {
-	return nil, nil
-}
+func (fakeHostService) GetInfraConfig() (*apiv1.InfraConfig, error) { return nil, nil }
```
jemoreira, line 121: most of these changes are no-ops and just introduce noise

jemoreira (approved):

pkg/cli/cli_test.go
ser-io: agree, I reverted them.

PR #300, opened 2024-04-02
Upload -imgs-.zip local artifact.

ser-io:


pkg/cli/cli.go
```diff
@@ -102,6 +102,7 @@ const (
 	localBootloaderSrcFlag    = "local_bootloader_src"
 	localCVDHostPkgSrcFlag    = "local_cvd_host_pkg_src"
 	localImagesSrcsFlag       = "local_images_srcs"
+	localImagesZipSrcFlag     = "local_images_zip_src"
```
jemoreira, line 105: is the user expected to create this zip file manually? or is it generated for them?

ser-io: good point, at the moment the flag value expects a single filename value, the file should have been already created.

jemoreira (approved):

PR #301, opened 2024-04-03
Implement http basic auth for cvdr.

denniscy1993:
Always permit users to access the homepage and the cloud_orchestrator config without specifying their username. Users can get the account manager type accordingly to decide whether they should set their username in following HTTP requests.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -22,6 +22,7 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"os/user"
 	"strings"
 	"time"
 
@@ -1194,16 +1195,33 @@ func buildServiceBuilder(builder client.ServiceBuilder, authnConfig *AuthnConfig
 			RetryDelay:     5 * time.Second,
 			ChunkSizeBytes: chunkSizeBytes,
 		}
-		if authnConfig != nil && authnConfig.OIDCToken != nil {
-			content, err := os.ReadFile(authnConfig.OIDCToken.TokenFile)
-			if err != nil {
-				return nil, fmt.Errorf("failed loading oidc token: %w", err)
+		if authnConfig != nil {
+			if authnConfig.OIDCToken != nil && authnConfig.HTTPBasicAuthn != nil {
+				return nil, fmt.Errorf("should only set one authentication option")
 			}
-			value := strings.TrimSuffix(string(content), "\n")
-			opts.Authn = &client.AuthnOpts{
-				OIDCToken: &client.OIDCToken{
+			opts.Authn = &client.AuthnOpts{}
+			if authnConfig.OIDCToken != nil {
+				content, err := os.ReadFile(authnConfig.OIDCToken.TokenFile)
+				if err != nil {
+					return nil, fmt.Errorf("failed loading oidc token: %w", err)
+				}
+				value := strings.TrimSuffix(string(content), "\n")
+				opts.Authn.OIDCToken = &client.OIDCToken{
 					Value: value,
-				},
+				}
+			} else if authnConfig.HTTPBasicAuthn != nil {
+				switch authnConfig.HTTPBasicAuthn.UsernameSrc {
+				case UnixUsernameSrc:
+					user, err := user.Current()
+					if err != nil {
+						return nil, fmt.Errorf("unable to get unix username for http basic authn: %w", err)
+					}
+					opts.Authn.HTTPBasic = &client.HTTPBasic{
+						Username: user.Username,
+					}
+				default:
+					return nil, fmt.Errorf("invalid http basic authn UsernameSrc type: %s", authnConfig.HTTPBasicAuthn.UsernameSrc)
+				}
 			}
 		}
 		return builder(opts)
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -55,10 +55,15 @@ func (e *ApiCallError) Is(target error) bool {
 
 type AuthnOpts struct {
 	OIDCToken *OIDCToken
+	HTTPBasic *HTTPBasic
 }
 
 type OIDCToken struct {
 	Value string
+}
+
+type HTTPBasic struct {
+	Username string
 }
 
 type ServiceOptions struct {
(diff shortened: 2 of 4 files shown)


ser-io:
Why users would need to check the CO config to set their username in follow up requests when the only information given when accessing the `/v1/config` endpoint right now is the instance manager type?

Can you given more context on the bigger problem we aim to start solving with this PR?

denniscy1993:
My goal is to enable `cvdr` when CO is using `HTTPBasicAccountManager`, and I think we should set the HTTP Basic Authentication information only for `HTTPBasicAccountManager` (more details here: b/332791987). I'll have a follow-up PR to add more information in `v1/config` so that `cvdr` can base on that information to decide whether to set the username or not.

jemoreira:
> so that cvdr can base on that information to decide whether to set the username or not.

Can `cvdr` read its own config file (`/etc/cvdr.toml`) to obtain that information? That's how it was originally designed.

0405ysj:
> Can cvdr read its own config file (/etc/cvdr.toml) to obtain that information? That's how it was originally designed.

Should we also consider the case that using multiple COs in the same time? For instance, a user want to access x86 and arm64 CF instances simultaneously with distinguished server have different AM type.

denniscy1993:
> > Can cvdr read its own config file (/etc/cvdr.toml) to obtain that information? That's how it was originally designed.
> 
> Should we also consider the case that using multiple COs in the same time? For instance, a user want to access x86 and arm64 CF instances simultaneously with distinguished server have different AM type.

By using the approach that @jemoreira suggested, I think we can set different `envVarUserConfigPath` to let `cvdr` talk to different servers. Did I understand it correctly?

0405ysj:
> By using the approach that @jemoreira suggested, I think we can set different envVarUserConfigPath to let cvdr talk to different servers.

Yes, it could when setting env variable, but I'm curious whether switching `cvdr.toml` file for using different COs is a good conceptual decision or not, since user wouldn't know any CO config info.(e.g. does `cvdr` user know what IM & AM type CO uses?)

denniscy1993:
Since our short-term goal is to let `cvdr` support HTTPBasicAM + DockerIM, I think we can use the `cvdr.toml` approach for this PR. Let's discuss more details at b/332791987.

> Yes, it could when setting env variable, but I'm curious whether switching `cvdr.toml` file for using different COs is a good conceptual decision or not, since user wouldn't know any CO config info I think. (e.g. does `cvdr` user know what IM & AM type CO uses?)

And regarding this, +1 to me that it would be useful to get the CO config via cvdr, so that I can check my local `cvdr.toml`. I didn't find such command right now (please correct me if I missed it). I can create another PR if we all agree that we want to support this feature.

ser-io:
supporting multiple CO deployments from `cvdr` needs a deeper discussion. I'll write something about this. Let's focus this PR on enabling the new parameter configuration as we were connecting to a CO deployment using HTTP Basic Authentication Account Manager type.

0405ysj:
> Let's focus this PR on enabling the new parameter configuration as we were connecting to a CO deployment using HTTP Basic Authentication Account Manager type.

Sounds good, and I agree that the higher priority should be supporting new AM type with docker IM type. It's enough for me that everyone is aware of needs to consider multiple COs someday.

ser-io (changes requested):
Let's have the commit message focused on cvdr only: i.e "Implement http basic auth for cvdr "

the "CO that uses HTTPBasicAccountManager" part is an implementation details which could be added in the detailed description of the commit message.

pkg/client/httputils.go
```diff
@@ -38,6 +38,8 @@ type HTTPHelper struct {
 	RootEndpoint string
 	Dumpster     io.Writer
 	AccessToken  string
+	UseHTTPBasic bool
```
ser-io, line 41: let's merge these two into `HTTPBasicUsername` and call `SetBasicAuth` if not empty.

pkg/client/client.go
```diff
@@ -54,13 +55,22 @@ func (e *ApiCallError) Is(target error) bool {
...
 
 type AuthnOpts struct {
-	OIDCToken *OIDCToken
+	OIDCToken      *OIDCToken
+	HTTPBasicAuthn *HTTPBasicAuthnConfig
 }
 
 type OIDCToken struct {
 	Value string
 }
```
ser-io, line 65: move the `HTTPBasicAuthnConfig` and the logic to get the username value into the config code in the `cli` package, the client code should receive the username value keeping the separation between the configuration logic and the http client logic.

We could have something like the following in the client package.

```
type AuthnOpts struct {
	OIDCToken *OIDCToken
	HTTPBasic *HTTPBasic
}

type HTTPBasic struct {
  Username string
}
```

pkg/client/httputils.go
```diff
@@ -164,6 +175,9 @@ func (rb *HTTPRequestBuilder) doWithRetries(retryOpts RetryOptions) (*http.Respo
 	if rb.helper.AccessToken != "" {
 		rb.AddHeader("Authorization", "Bearer "+rb.helper.AccessToken)
 	}
+	if rb.helper.UseHTTPBasic {
```
ser-io, line 178: return err if both `AccessToken` and `HTTPBasicUsername` are set.

ser-io (changes requested):

pkg/cli/cli.go
```diff
@@ -1194,16 +1195,31 @@ func buildServiceBuilder(builder client.ServiceBuilder, authnConfig *AuthnConfig
...
-			content, err := os.ReadFile(authnConfig.OIDCToken.TokenFile)
-			if err != nil {
-				return nil, fmt.Errorf("failed loading oidc token: %w", err)
-			}
-			value := strings.TrimSuffix(string(content), "\n")
-			opts.Authn = &client.AuthnOpts{
-				OIDCToken: &client.OIDCToken{
+		if authnConfig != nil {
+			opts.Authn = &client.AuthnOpts{}
+			if authnConfig.OIDCToken != nil {
```
ser-io, line 1200: we can add validation here as well to make sure only one authentication option is used in the configuration.

pkg/cli/cli.go
```diff
@@ -1194,16 +1195,31 @@ func buildServiceBuilder(builder client.ServiceBuilder, authnConfig *AuthnConfig
...
+				if err != nil {
+					return nil, fmt.Errorf("failed loading oidc token: %w", err)
+				}
+				value := strings.TrimSuffix(string(content), "\n")
+				opts.Authn.OIDCToken = &client.OIDCToken{
 					Value: value,
-				},
+				}
+			}
+			if authnConfig.HTTPBasicAuthn != nil {
```
ser-io, line 1210: nit: use `else if` to double down on having only one authentication option set.

ser-io (approved):

pkg/cli/cli.go
denniscy1993: Done.

pkg/cli/cli.go
denniscy1993: Done, and I also did it at `pkg/client/httputils.go`.

ikicha (approved):

0405ysj (approved):

PR #302, opened 2024-04-15
Prompt the client to set an username for http basic auth.

denniscy1993:
- The cloud orchestrator server that uses HTTPBasicAccountManager would send a 401 unauthorized response to the client to popup the login window.

### DIFF
--- a/pkg/app/accounts/basic.go
+++ b/pkg/app/accounts/basic.go
@@ -16,8 +16,6 @@ package accounts
 
 import (
 	"net/http"
-
-	apperr "github.com/google/cloud-android-orchestration/pkg/app/errors"
 )
 
 const HTTPBasicAMType AMType = "http-basic"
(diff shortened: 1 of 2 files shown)


denniscy1993:
This PR is to address b/333545755.

pkg/app/accounts/basic.go
```diff
@@ -32,7 +30,11 @@ func NewHTTPBasicAccountManager() *HTTPBasicAccountManager {
 }
 
 func (m *HTTPBasicAccountManager) UserFromRequest(r *http.Request) (User, error) {
-	return userFromRequest(r)
+	username, _, ok := r.BasicAuth()
+	if !ok {
+		return nil, nil
```
0405ysj, line 35: nit: Don't we need an error here? This seems like a refactoring, but error looks gone.

0405ysj (approved):

pkg/app/accounts/basic.go
denniscy1993: Actually `app.go` will send `apperr.NewUnauthenticatedError("Authentication required", nil)` when it receives  a `nil` user returned by  `UserFromRequest`. We need this 401 unauthorized error to popup the login window at the client side.

ser-io: basic auth is considered a bad practice, it could work in a very controlled environment as a POC, which is the use case here, having said so, we should never prompt the user for a password.  please if this popup have password input, let's discard this PR as users can mistakenly enter a valid password which is not wanted or needed in our system.

ser-io (changes requested):
we should not prompt users for a password, it's not needed and users can mistakenly provide one.

denniscy1993:
Thanks for the suggestions, I would close this PR.

PR #305, opened 2024-04-23
Change host address of docker instance

0405ysj:
This PR is just for refactoring and optimization, but following change will be able to grab IP address of docker instance, to do adb connection via cvdr.

### DIFF
--- a/pkg/app/instances/docker.go
+++ b/pkg/app/instances/docker.go
@@ -223,15 +223,11 @@ func (m *DockerInstanceManager) WaitOperation(zone string, _ accounts.User, name
 	}
 }
 
-func (m *DockerInstanceManager) GetHostAddr() (string, error) {
-	return "127.0.0.1", nil
-}
-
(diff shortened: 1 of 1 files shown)


pkg/app/instances/docker.go
```diff
@@ -82,7 +82,7 @@ func (m *DockerInstanceManager) CreateHost(zone string, _ *apiv1.CreateHostReque
 		},
 	}
 	hostConfig := &container.HostConfig{
-		Privileged:      true,
+		Privileged: true,
```
ser-io, line 85: nit: amend this change to the previous commit, having 2 commits on this PR only.

ser-io (approved):

ikicha (approved):

pkg/app/instances/docker.go
0405ysj: Amended formatting change into previous one, Done.

PR #307, opened 2024-04-24
Introduce --connect_agent with the new type(proxy_agent) into cvdr connect

0405ysj:
How to turn on `--connect_agent` flag with `proxy_agent` in `cvdr connect`:

```
./cvdr \
--http_proxy=socks5://localhost:1337 \
--service_url=http://100.114.242.57:8080 \
--zone=local \
--host=$CO_HOST_NAME \
--connect_agent=proxy_agent \
connect
```

Currently this PR needs some fixes:
- Agent process should be terminated(Make it being daemonized): WIP

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -20,10 +20,14 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"net"
 	"net/http"
+	"net/url"
 	"os"
+	"os/signal"
 	"os/user"
 	"strings"
+	"syscall"
 	"time"
 
 	client "github.com/google/cloud-android-orchestration/pkg/client"
@@ -32,6 +36,7 @@ import (
 	"github.com/PaesslerAG/jsonpath"
 	"github.com/hashicorp/go-multierror"
 	"github.com/spf13/cobra"
+	"golang.org/x/net/proxy"
 	"golang.org/x/term"
 )
 
@@ -107,9 +112,10 @@ const (
 )
 
 const (
-	ConnectCommandName         = "connect"
-	DisconnectCommandName      = "disconnect"
-	ConnectionAgentCommandName = "agent"
+	ConnectCommandName               = "connect"
+	DisconnectCommandName            = "disconnect"
+	ConnectionWebRTCAgentCommandName = "webrtc_agent"
+	ConnectionProxyAgentCommandName  = "proxy_agent"
 )
 
 const (
@@ -179,7 +185,8 @@ type ConnectFlags struct {
 	host             string
 	skipConfirmation bool
 	// Path to file containing the ICE configuration to be used in the underlaying WebRTC connection.
-	ice_config string
+	ice_config   string
+	connectAgent string
 }
 
 func (f *ConnectFlags) AsArgs() []string {
@@ -584,7 +591,7 @@ func cvdCommands(opts *subCommandOpts) []*cobra.Command {
 }
 
 func connectionCommands(opts *subCommandOpts) []*cobra.Command {
-	connFlags := &ConnectFlags{CVDRemoteFlags: opts.RootFlags, host: "", skipConfirmation: false}
+	connFlags := &ConnectFlags{CVDRemoteFlags: opts.RootFlags, host: "", skipConfirmation: false, connectAgent: ConnectionWebRTCAgentCommandName}
 	connect := &cobra.Command{
 		Use:   ConnectCommandName,
 		Short: "(Re)Connects to a CVD and tunnels ADB messages",
@@ -596,6 +603,7 @@ func connectionCommands(opts *subCommandOpts) []*cobra.Command {
 	connect.Flags().BoolVarP(&connFlags.skipConfirmation, "yes", "y", false,
 		"Don't ask for confirmation for closing multiple connections.")
 	connect.Flags().StringVar(&connFlags.ice_config, iceConfigFlag, "", iceConfigFlagDesc)
+	connect.Flags().StringVar(&connFlags.connectAgent, "connect_agent", ConnectionWebRTCAgentCommandName, "Connect agent type")
 	disconnect := &cobra.Command{
 		Use:   fmt.Sprintf("%s <foo> <bar> <baz>", DisconnectCommandName),
 		Short: "Disconnect (ADB) from CVD",
@@ -606,17 +614,26 @@ func connectionCommands(opts *subCommandOpts) []*cobra.Command {
 	disconnect.Flags().StringVar(&connFlags.host, hostFlag, "", "Specifies the host")
 	disconnect.Flags().BoolVarP(&connFlags.skipConfirmation, "yes", "y", false,
 		"Don't ask for confirmation for closing multiple connections.")
-	agent := &cobra.Command{
+	webrtcAgent := &cobra.Command{
 		Hidden: true,
-		Use:    ConnectionAgentCommandName,
+		Use:    ConnectionWebRTCAgentCommandName,
 		RunE: func(c *cobra.Command, args []string) error {
-			return runConnectionAgentCommand(connFlags, &command{c, &connFlags.Verbose}, args, opts)
+			return runConnectionWebrtcAgentCommand(connFlags, &command{c, &connFlags.Verbose}, args, opts)
 		},
 	}
-	agent.Flags().StringVar(&connFlags.host, hostFlag, "", "Specifies the host")
-	agent.Flags().StringVar(&connFlags.ice_config, iceConfigFlag, "", iceConfigFlagDesc)
-	agent.MarkPersistentFlagRequired(hostFlag)
-	return []*cobra.Command{connect, disconnect, agent}
+	webrtcAgent.Flags().StringVar(&connFlags.host, hostFlag, "", "Specifies the host")
+	webrtcAgent.Flags().StringVar(&connFlags.ice_config, iceConfigFlag, "", iceConfigFlagDesc)
+	webrtcAgent.MarkPersistentFlagRequired(hostFlag)
+	proxyAgent := &cobra.Command{
+		Hidden: true,
+		Use:    ConnectionProxyAgentCommandName,
+		RunE: func(c *cobra.Command, args []string) error {
+			return runConnectionProxyAgentCommand(connFlags, &command{c, &connFlags.Verbose}, args, opts)
+		},
+	}
+	proxyAgent.Flags().StringVar(&connFlags.host, hostFlag, "", "Specifies the host")
+	proxyAgent.MarkPersistentFlagRequired(hostFlag)
+	return []*cobra.Command{connect, disconnect, webrtcAgent, proxyAgent}
 }
 
 func runCreateHostCommand(c *cobra.Command, flags *CreateHostFlags, opts *subCommandOpts) error {
@@ -731,7 +748,7 @@ func runCreateCVDCommand(c *cobra.Command, args []string, flags *CreateCVDFlags,
 	if flags.CreateCVDOpts.AutoConnect {
 		for _, cvd := range cvds {
 			statePrinter.Print(fmt.Sprintf(connectCVDStateMsgFmt, cvd.WebRTCDeviceID))
-			cvd.ConnStatus, err = ConnectDevice(flags.CreateCVDOpts.Host, cvd.WebRTCDeviceID, "", &command{c, &flags.Verbose}, opts)
+			cvd.ConnStatus, err = ConnectDevice(flags.CreateCVDOpts.Host, cvd.WebRTCDeviceID, "", ConnectionWebRTCAgentCommandName, &command{c, &flags.Verbose}, opts)
 			statePrinter.PrintDone(fmt.Sprintf(connectCVDStateMsgFmt, cvd.WebRTCDeviceID), err)
 			if err != nil {
 				merr = multierror.Append(merr, fmt.Errorf("failed to connect to device: %w", err))
@@ -840,7 +857,7 @@ func promptHostNameSelection(c *command, service client.Service, selOpt Selectio
 
 // Starts a connection agent process and waits for it to report the connection was
 // successfully created or an error occurred.
-func ConnectDevice(host, device, ice_config string, c *command, opts *subCommandOpts) (*ConnStatus, error) {
+func ConnectDevice(host, device, ice_config, agent string, c *command, opts *subCommandOpts) (*ConnStatus, error) {
 	// Clean old logs files as we are about to create new ones.
 	go func() {
 		minAge := opts.InitialConfig.LogFilesDeleteThreshold()
@@ -857,7 +874,7 @@ func ConnectDevice(host, device, ice_config string, c *command, opts *subCommand
 		host:           host,
 		ice_config:     ice_config,
 	}
-	cmdArgs := buildAgentCmdArgs(flags, device)
+	cmdArgs := buildAgentCmdArgs(flags, device, agent)
 
 	output, err := opts.CommandRunner.StartBgCommand(cmdArgs...)
 	if err != nil {
@@ -944,7 +961,7 @@ func runConnectCommand(flags *ConnectFlags, c *command, args []string, opts *sub
 		go func(connCh chan ConnStatus, errCh chan error, cvd RemoteCVDLocator) {
 			defer close(connCh)
 			defer close(errCh)
-			status, err := ConnectDevice(cvd.Host, cvd.WebRTCDeviceID, flags.ice_config, c, opts)
+			status, err := ConnectDevice(cvd.Host, cvd.WebRTCDeviceID, flags.ice_config, flags.connectAgent, c, opts)
 			if err != nil {
 				errCh <- fmt.Errorf("failed to connect to %q on %q: %w", cvd.WebRTCDeviceID, cvd.Host, err)
 			} else {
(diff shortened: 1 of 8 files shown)


pkg/cli/cli.go
```diff
@@ -596,6 +603,8 @@ func connectionCommands(opts *subCommandOpts) []*cobra.Command {
 	connect.Flags().BoolVarP(&connFlags.skipConfirmation, "yes", "y", false,
 		"Don't ask for confirmation for closing multiple connections.")
 	connect.Flags().StringVar(&connFlags.ice_config, iceConfigFlag, "", iceConfigFlagDesc)
+	connect.Flags().BoolVar(&connFlags.doNotUseWebrtc, "do_not_use_webrtc", false,
```
jemoreira, line 606: `do_not_use_webrtc` is a mouthful, maybe this could be `--webrtc` with default to true, and a complementary one like `--ssh_port_forwarding`.

Alternatively, this could be a single flag like `--transport=webrtc|ssh|...`

ikicha: +1 for @jemoreira 's comment. it would be --transport=port_forwarding or something(because it doesn't use ssh directly)

ikicha: cvdr does port forwarding between docker ip:<adb port> and local unix domain socket file(like `socat UNIX-LISTEN:/tmp/my.sock,fork TCP:<docker internal ip>:6520`) via socks5 proxy

jemoreira: I see, then `port_forwarding` or even leaving it empty are better than my suggestion.

0405ysj: I understood your concern. At least, I won't set naming of flag as `do_not_use_webrtc`.

However, I think neither `ssh` nor `port_forwarding` is the good naming of this because:
- We don't use `ssh` when not using SOCKS5 proxy
- I don't think we can say this as port forwarding, since an endpoint is not port, just a file-based socket.

So I just want to set the name as `proxy`. However if you have any concern on the naming, please let me know :) I think changing it again would be very simple task.

0405ysj: Done for this. I'll resolve this conversation tomorrow.

0405ysj:
Though current commit requires implementation about daemonization/disconnection/re-connection yet,  this PR can make `cvdr connect` possible.

0405ysj:
A dumb question, which would you prefer for the creation of daemon process(using functions; `exec.Command` and `Cmd.start`) to keep socket connection?
- Let background command (proxy_agent) creates another daemon process.
- Let background command (proxy_agent) be daemon process.

At first, I thought a way using `fork`-like functions, but seems it's not supported in golang itself or other 3p libraries. (At least I checked https://github.com/sevlyar/go-daemon)

0405ysj:
> Let background command (proxy_agent) be daemon process.

If there's any existing preferences, I'll keep trying working with this.

0405ysj:
I changed the code as daemonized, but WIP for disconnection, re-connection and listing existing connection.
I think it requires a control socket for terminating process successfully but code regarding control socket seems highly coupled with webrtc stuff.

pkg/cli/cli.go
```diff
@@ -983,14 +1001,99 @@ func printConnection(c *command, cvd RemoteCVDLocator, status ConnStatus) {
...
-	args := []string{
-		ConnectionAgentCommandName,
-		device,
-	}
+func buildAgentCmdArgs(flags *ConnectFlags, device string, connectAgent string) []string {
+	args := []string{connectAgent, device}
 	return append(args, flags.AsArgs()...)
 }
 
+func runConnectionProxyAgentCommand(flags *ConnectFlags, c *command, args []string, opts *subCommandOpts) error {
```
ser-io, line 1043: let's add similar documentation as in `runConnectionWebrtcAgentCommand`

pkg/cli/cli.go
```diff
@@ -983,14 +1001,99 @@ func printConnection(c *command, cvd RemoteCVDLocator, status ConnStatus) {
...
+		return errors.New("instance type should be Docker")
+	}
+
+	controlDir := opts.InitialConfig.ConnectionControlDirExpanded()
+	cvdHosts, err := listCVDsSingleHost(service, controlDir, flags.host)
+	if err != nil {
+		return fmt.Errorf("error listing CVDs: %w", err)
+	}
+	var port string
+	for _, cvd := range cvdHosts[0].CVDs {
```
ser-io, line 1078: to improve readability, let's split this logic into two parts:

1. find cvd
2. get port value.

something like.
```
cvd := findCVD(host, name)
_, port, err := net.SplitHostPort(cvd.ADBSerial)
```

pkg/cli/cli.go
```diff
@@ -983,14 +1001,99 @@ func printConnection(c *command, cvd RemoteCVDLocator, status ConnStatus) {
...
+	}
+	socketName := strings.Join([]string{"proxy", shortenHost, device}, "_") + ".sock"
+	localListener, err := net.Listen("unix", filepath.Join(controlDir, socketName))
+	if err != nil {
+		return fmt.Errorf("failed to create unix socket: %w", err)
+	}
+	defer localListener.Close()
+	localConn, _ := localListener.Accept()
+
+	go io.Copy(remoteConn, localConn)
```
ser-io, line 1120: handle errors here.

pkg/cli/cli.go
```diff
@@ -983,14 +1001,99 @@ func printConnection(c *command, cvd RemoteCVDLocator, status ConnStatus) {
...
+	if err != nil {
+		return fmt.Errorf("failed to dial remote port: %w", err)
+	}
+	defer remoteConn.Close()
+
+	shortenHost := flags.host
+	if len(shortenHost) >= 8 {
+		shortenHost = shortenHost[:8]
+	}
+	socketName := strings.Join([]string{"proxy", shortenHost, device}, "_") + ".sock"
```
ser-io: let's move the logic to build to socket name into a helper function. use `fmt.Sprintf()`

pkg/cli/cli.go
0405ysj: Done.

pkg/cli/cli.go
```diff
@@ -983,14 +1001,99 @@ func printConnection(c *command, cvd RemoteCVDLocator, status ConnStatus) {
...
 
+func runConnectionProxyAgentCommand(flags *ConnectFlags, c *command, args []string, opts *subCommandOpts) error {
+	if len(args) > 1 {
+		return fmt.Errorf("connection agent only supports a single device, received: %v", args)
+	}
+	if len(args) == 0 {
+		return errors.New("missing device")
+	}
+	device := args[0]
+	service, err := opts.ServiceBuilder(flags.CVDRemoteFlags, c.Command)
```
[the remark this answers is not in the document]
0405ysj: Added a new function `findHost`.

pkg/cli/cli.go
0405ysj: Done.

pkg/cli/cli.go
0405ysj: Done.

pkg/cli/cli.go
0405ysj: Done.

pkg/cli/cli.go
```diff
@@ -983,23 +1001,165 @@ func printConnection(c *command, cvd RemoteCVDLocator, status ConnStatus) {
...
-		device,
-	}
+func buildAgentCmdArgs(flags *ConnectFlags, device string, connectAgent string) []string {
+	args := []string{connectAgent, device}
 	return append(args, flags.AsArgs()...)
 }
 
-// Handler for the agent command. This is not meant to be called by the user
-// directly, but instead is started by the open command.
+func enableAdbConnection(socketPath string, enable bool) error {
```
ikicha, line 1009: can we use https://github.com/google/cloud-android-orchestration/blob/main/pkg/cli/adbserver.go? (we need to modify it to accept localfilesystem: though)

pkg/cli/cli.go
```diff
@@ -983,23 +1001,165 @@ func printConnection(c *command, cvd RemoteCVDLocator, status ConnStatus) {
...
+		return fmt.Errorf("failed to find cvd: %w", err)
+	}
+	_, port, err := net.SplitHostPort(cvd.ADBSerial)
+	if err != nil {
+		return fmt.Errorf("failed to parse port from ADB serial: %w", err)
+	}
+	if port == "" {
+		return errors.New("failed to find port")
+	}
+
```
ikicha, line 1060: how about extracting this block to a separate method like findDeviceAddress or something similar?

pkg/cli/cli.go
```diff
@@ -983,23 +1001,165 @@ func printConnection(c *command, cvd RemoteCVDLocator, status ConnStatus) {
...
+		if _, err := io.Copy(remoteConn, localConn); err != nil {
+			errorCh <- fmt.Errorf("no longer able to copy data from local to remote: %w", err)
+		}
+	}()
+	go func() {
+		localConn := <-startRemoteToLocalCh
+		if _, err := io.Copy(localConn, remoteConn); err != nil {
+			errorCh <- fmt.Errorf("no longer able to copy data from remote to local: %w", err)
+		}
+	}()
```
ikicha, line 1130: it would be another separate method as well(forward?)

pkg/cli/cli.go
0405ysj: Done.

pkg/cli/cli.go
0405ysj: Done.

pkg/cli/cli.go
0405ysj: We can use it, done.

ikicha (approved):
LGTM, thanks!

0405ysj:
Current approach is making `proxy_agent` as non-daemon process, not relying on the control socket.
Pressing `ctrl+c`(SIGINT) to running `cvdr connect` process will stop ADB connection.

0405ysj:
@ser-io @jemoreira Would you want to leave any comments here? I'm having a plan to merge this tomorrow morning.

pkg/cli/host.go
```diff
@@ -51,3 +53,16 @@ func hostnames(service client.Service) ([]string, error) {
...
+	hosts, err := service.ListHosts()
+	if err != nil {
+		return nil, fmt.Errorf("error listing hosts: %w", err)
+	}
+	for _, host := range hosts.Items {
+		if host.Name == hostname {
+			return host, nil
+		}
+	}
+	return nil, fmt.Errorf("no such host: %s", hostname)
```
ser-io, line 67: nit: `name not found: %s`.

pkg/cli/host.go
```diff
@@ -51,3 +53,16 @@ func hostnames(service client.Service) ([]string, error) {
 	}
 	return result, nil
 }
+
+func findHost(service client.Service, hostname string) (*apiv1.HostInstance, error) {
```
ser-io, line 57: nit: `name` only is ok here. (https://google.github.io/styleguide/go/decisions.html#variable-names)

pkg/cli/cli.go
```diff
@@ -983,23 +1000,166 @@ func printConnection(c *command, cvd RemoteCVDLocator, status ConnStatus) {
...
+
+	// Wait for channels notifying the termination.
+	startTermination := false
+	var chErr error
+	for !startTermination {
```
ser-io, line 1111: we expect only one signal on either `case` branch (whichever comes first) in order to continue. I think the outer for loop might not be necessary we are doing just one iteration.

pkg/cli/cli.go
```diff
@@ -983,23 +1000,166 @@ func printConnection(c *command, cvd RemoteCVDLocator, status ConnStatus) {
...
+		select {
+		case <-sigCh:
+			startTermination = true
+		case chErr = <-errorCh:
+			startTermination = true
+		}
+	}
+
+	// Remove the socket file and disconnect ADB before the termination of this process.
+	adbServerProxy.DisconnectWithLocalFileSystem(socketPath)
```
ser-io, line 1121: should we have this cleanup in a `defer` call?

pkg/cli/cli.go
```diff
@@ -983,23 +1000,166 @@ func printConnection(c *command, cvd RemoteCVDLocator, status ConnStatus) {
...
+		if err := adbServerProxy.ConnectWithLocalFileSystem(socketPath); err != nil {
+			errorCh <- fmt.Errorf("failed to connect adb: %w", err)
+		}
+	}()
+
+	// Listener is accepted by ADB connection.
+	startLocalToRemoteCh := make(chan net.Conn)
+	startRemoteToLocalCh := make(chan net.Conn)
+	go func() {
+		localConn, err := localListener.Accept()
```
ser-io, line 1084: wondering whether the `Accept` call can be done in the same `goroutine` to avoid the `startLocalToRemoteCh` and `startRemoteToLocalCh` channels.

ser-io (changes requested):

pkg/cli/cli.go
```diff
@@ -107,9 +112,10 @@ const (
 )
 
 const (
-	ConnectCommandName         = "connect"
-	DisconnectCommandName      = "disconnect"
-	ConnectionAgentCommandName = "agent"
+	ConnectCommandName               = "connect"
+	DisconnectCommandName            = "disconnect"
+	ConnectionWebrtcAgentCommandName = "webrtc_agent"
```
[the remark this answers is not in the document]
0405ysj: Done.

pkg/cli/cli.go
```diff
@@ -840,7 +857,7 @@ func promptHostNameSelection(c *command, service client.Service, selOpt Selectio
 
 // Starts a connection agent process and waits for it to report the connection was
 // successfully created or an error occurred.
-func ConnectDevice(host, device, ice_config string, c *command, opts *subCommandOpts) (*ConnStatus, error) {
+func ConnectDevice(host, device, ice_config string, connectAgent string, c *command, opts *subCommandOpts) (*ConnStatus, error) {
```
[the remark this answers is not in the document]
0405ysj: Done.

pkg/cli/host.go
0405ysj: Done.

pkg/cli/host.go
0405ysj: Done.

pkg/cli/conn.go
```diff
@@ -551,6 +556,20 @@ func logsDir(controlDir string) string {
 	return controlDir + "/logs"
 }
 
+func proxyDir(controlDir string) string {
+	return controlDir + "/proxy"
```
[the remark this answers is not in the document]
0405ysj: Done.

pkg/cli/cli.go
```diff
@@ -983,23 +1000,166 @@ func printConnection(c *command, cvd RemoteCVDLocator, status ConnStatus) {
...
-	}
+func buildAgentCmdArgs(flags *ConnectFlags, device string, connectAgent string) []string {
+	args := []string{connectAgent, device}
 	return append(args, flags.AsArgs()...)
 }
 
-// Handler for the agent command. This is not meant to be called by the user
-// directly, but instead is started by the open command.
+// Return the ADB address of remote Cuttlefish instance as [host]:port.
+func findDeviceAdbAddress(service client.Service, controlDir, hostname, device string) (string, error) {
```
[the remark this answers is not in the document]
0405ysj: Removed the logic introducing the new function`findDeviceAdbAddress`. OTOH, I thought introducing `adbPort` would't worth that much as well, since `adbPort(cvd *CVD)` will contain `net.SplitHostPort` only inside. I picked it up all logics into `runConnectionProxyAgentCommand` like before.

pkg/cli/cli.go
0405ysj: I think I missed this one during rewriting from old patchset. Thanks for mentioning! Done.

pkg/cli/cli.go
0405ysj: Current execution flow is,
1. Start running `adbServerProxy.ConnectWithLocalFileSystem(socketPath)`.
2. Blocked by `localListener.Accept()` is disappeared.
3. Start running `io.copy` for both direction.
4. Finish running `adbServerProxy.ConnectWithLocalFileSystem(socketPath)`.

The problem is, both `io.copy` are intended to run forever while keeping the ADB connection. So each `io.copy` needs its own goroutine. We can reduce one channel, but we cannot remove both connection. But in the perspective of readability, I think reducing only one channel wouldn't be the better one. Please reopen this if you have another idea.

pkg/cli/cli.go
0405ysj: Done for ADB disconnection. For removing socket, we may lose propagating unexpected error that socket is not able to remove, so I kept.

pkg/cli/cli.go
ser-io: Sorry, I miscommunicated. 

I wanted to know whether we could get rid of the goroutine for `localConn, err := localListener.Accept()`, this way the `localConn` could be used in the following `io.Copy` goroutines removing the need of having two different channels to send the `localConn` instance over.

If we do need the `goroutine`, can we have one channel (`localConnCh`)  and send the `localConn` instance twice?

ser-io (changes requested):

pkg/cli/cli.go
0405ysj: IMO, we cannot remove the goroutine for `localListener.Accept()`. `. While waiting for `localListener.Accept()` w/o goroutine, we cannot wait `errCh` sent from `adbServerProxy.ConnectWithLocalFileSystem(socketPath)`.

And about having one channel(`localConnCh`), I already tried it before. However in that trial, one of 2 goroutines couldn't get `localConn` properly. That's why I defined 2 channels for transmitting `localConn`.

ser-io (approved):

PR #308, opened 2024-05-08
Update documentation to run cloud orchestrator with dockerIMType.

0405ysj:
Precondition for this PR: https://github.com/google/android-cuttlefish/pull/549


docs/docker.md
```diff
@@ -0,0 +1,160 @@
+# Prepare docker image
+
+Please follow the docker part of
+[README.md](https://github.com/google/android-cuttlefish/blob/main/README.md#docker)
+in `google/android-cuttlefish` github repository, and check if
+`cuttlefish-linux-infrastructure` exists.
```
denniscy1993, line 6: should we explain how to check this?

cloud_orchestrator_config_docker.toml
denniscy1993, line 1: how about putting this file under a sub-directory? For example, "examples" or "tests".

docs/docker.md
0405ysj: That's why the precondition of this PR is https://github.com/google/android-cuttlefish/pull/549. Sorry for not pushing second commit at there. README.md at there will explain how to check.

cloud_orchestrator_config_docker.toml
0405ysj: I think it should be considered with where to put `secrets.json` file as well. In perspective of putting config file under sub-dir, I think `secrets.json` should be in sub-dir as well. However, config file should have information about `secrets.json` located at sub-dir as well.

@jemoreira @ser-io About `secrets.json`, I have 2 questions:
- What `secrets.json` stand for? Do we really require `secrets.json` for UNIX SecretManager type?
- Should `secrets.json` be located at the parent dir, based on the information in `/path/to/cloud-android-orchestration/conf.toml`? I think the only use case for using `secrets.json` would be LocalIMType or DockerIMType, correct?

ser-io:
let's follow the same pattern we have to deploy to App Engine here, see https://github.com/google/cloud-android-orchestration/tree/main/scripts/gcp/app, resulting in something like the following:

```
scrips/
  gcp/
    ...
  docker/
    conf.toml
    build.sh|run.sh
```

Now in the repo main README.md we can have a new section specifically for docker where we can simply say something around: "Run ./scrips/docker/run.sh"

The steps listed in the new [docs/docker.md](https://github.com/google/cloud-android-orchestration/pull/308/commits/e1099ca9d43868a0922b91841200a6850c35cb53#diff-ae5611658dcd443d7a30575bebcdf15a71b7a0f5abebb18b62d00aa0373cae74) file should be `scrips/docker/build.sh`.

cloud_orchestrator_config_docker.toml
```diff
@@ -0,0 +1,36 @@
+WebStaticFilesPath = "web"
```
ser-io, line 1: let's move this into the new subdirectory `scrips/docker`

docs/docker.md
```diff
@@ -0,0 +1,169 @@
+# Prepare docker image
```
ser-io, line 1: make the steps here a bash scripts under `scrips/docker/`

docs/docker.md
```diff
@@ -62,3 +62,108 @@ If there's a network or security issue for accessing cloud orchestrator
...
+```
+
+## Use cvdr step-by-step
+
+### Host(docker instance) creation
+
+Please run the command below,
+```bash
+./cvdr \
+--service_url=http://${cloud_orchestrator_ip_address}:8080 \
```
ser-io, line 83: create a `SERVICE_URL` env variable. `SERVICE_URL="http://${cloud_orchestrator_ip_address}:8080"`

secrets_docker.json
```diff
@@ -0,0 +1,4 @@
+{
```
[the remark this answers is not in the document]
0405ysj: Shouldn't we put secrets.json file itself into `scripts/docker`, not generating by the build script?

docs/docker.md
0405ysj: A tricky point to construct bash script under `scripts/docker` is, we should do float value comparing for DOCKER_API_VERSION.

- To sync docker API version between internal and external code, the docker API version in the code level should be 1.43.
- Some of our ARM server installed docker package with API version as 1.41. Note that it's hard to upgrade into 1.43.
- We could fix DOCKER_API_VERSION as 1.41, but we would lose capability to use docker-based CO if the machine installed docker lower than 1.41.

0405ysj: Self-answer: seems `sort` function have version sort as its flag.

cloud_orchestrator_config_docker.toml
0405ysj: Put in `scripts/docker`.

secrets_docker.json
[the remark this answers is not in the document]
0405ysj: Done with putting `secrets.json` into `scripts/docker`, since it seems better.

docs/cvdr.md
```diff
@@ -0,0 +1,112 @@
...
+Please run:
+```bash
+cd /path/to/cloud-android-orchestration
+go build ./cmd/cvdr
+```
+
+## Find URL of cloud orchestrator
+
+If cloud orchestrator manages docker instances, please read
+[docker.md](docker.md) to check URL of cloud orchestrator.
```
denniscy1993, line 16: We may also describe how to get the `SERVICE_URL` when the CO is not using docker IM. Is it `http://${CLOUD_ORCHESTRATOR_IP_ADDRESS}`?

scripts/docker/run.sh
```diff
@@ -0,0 +1,35 @@
...
+docker_api_version_code=1.43
+docker_api_version_installed=$(docker version --format '{{.Client.APIVersion}}')
+docker_api_version=$(\
+  echo -e $docker_api_version_code\\n$docker_api_version_installed \
+  | sort --version-sort \
+  | head -n 1)
+
+dir=$(dirname $0)
+pushd $dir > /dev/null
+
```
denniscy1993, line 29: just curious: why we need this?

It seems to me that we can successfully get `cloud_android_orchestration_root` anywhere under the cloud-android-orchestration checkout directory.

scripts/docker/build.sh
```diff
@@ -0,0 +1,29 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Shell script for building Cloud Orchestrator
+
+dir=$(dirname $0)
+pushd $dir > /dev/null
```
denniscy1993, line 20: same as another comment, I am wondering if we really need this.

scripts/docker/build.sh
denniscy1993, line 1: This seems to be a general script. How about moving to the parent directory? (i.e. scripts/)

docs/cvdr.md
0405ysj: It would be, but there's no external use cases of `GCEIMType` currently AFAIK.

For local cases, the service url should be `http://localhost:8080`.

scripts/docker/run.sh
0405ysj: I added that because of a small difference. I thought that one would make the shell script more stable.

- Without `pushd $(dirname $0)`, `cloud_android_orchestration_root` is findable when current dir is under `cloud-android-orchestration` directory.
- With `pushd $(dirname $0)`, `cloud_android_orchestration_root` is findable even the outside of `cloud-android-orchestration` directory.

scripts/docker/build.sh
0405ysj: ditto.

scripts/docker/build.sh
0405ysj: That sounds good.

scripts/docker/run.sh
denniscy1993: I see. That's a good point!

docs/docker.md
```diff
@@ -0,0 +1,45 @@
...
+Please follow the docker part of
+[README.md](https://github.com/google/android-cuttlefish/blob/main/README.md#docker)
+in `google/android-cuttlefish` github repository, and check if the docker image
+`cuttlefish-orchestration` exists.
+
+## Build cloud orchestrator
+
+To build cloud orchestrator, please execute:
+```bash
+/path/to/cloud-android-orchestration/scripts/docker/build.sh
```
ser-io, line 17: avoid the `/path/to/cloud-android-orchestration` prefix using  `git clone https://github.com/google/cloud-android-orchestration.git clone && cd cloud-android-orchestration` as the first step, now the reader understand we are within the repository directory.

docs/docker.md
```diff
@@ -0,0 +1,45 @@
...
+containing Host Orchestrator inside.
+
+## Prepare docker image
+
+Please follow the docker part of
+[README.md](https://github.com/google/android-cuttlefish/blob/main/README.md#docker)
+in `google/android-cuttlefish` github repository, and check if the docker image
+`cuttlefish-orchestration` exists.
+
+## Build cloud orchestrator
```
ser-io, line 13: there's nothing specific when building the binary in the context of docker, it's built as always: `go build ./cmd/cloud_orchestrator`. This step in the documentation as well as the standalone script [scripts/docker/build.sh](https://github.com/google/cloud-android-orchestration/pull/308/commits/c349ce331a5f358b8aec4ad252d8e0b2def630e0#diff-be45b1c53fa8ef7864c013a77325ed29aca3eeae2be39a37e107d63e0d6134d1) makes it seems that we are building for docker in an special way. 

What's is definitely specific to docker is running the binary! In this case, my suggestion is too keep only the `run.sh` script  and make the build command `go build ./cmd/cloud_orchestrator` part of it, then remove the `build.sh` script and the documentation about this step.

scripts/docker/conf.toml
```diff
@@ -0,0 +1,36 @@
...
+Type = "unix"
+
+[AccountManager.OAuth2]
+Provider = "Google"
+RedirectURL = "http://localhost:8080/oauth2callback"
+
+[SecretManager]
+Type = ""
+
+[SecretManager.UNIX]
```
ser-io, line 17: I think this is not needed, you are using the empty Secret Manager Type: https://github.com/google/cloud-android-orchestration/blob/main/pkg/app/secrets/empty.go so there's no reason to define `SecretFilePath`. do you have any specific error if you omit this entry?

secrets_docker.json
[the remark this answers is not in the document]
ser-io: the config entry `[SecretManager.UNIX]` is not needed as you are using the empty secret manager, this way there's no need to have a `secrets.json` file.

docs/docker.md
0405ysj: Done.

scripts/docker/conf.toml
0405ysj: No, I just had no idea about treating `SecretManager`, since existing one(`/conf.toml`) does. Removed `secrets.json`.

By the way, what's the purpose for keeping conf.toml in the root directory of this repository?

docs/docker.md
0405ysj: I moved it from `scripts/docker/build.sh` to `scripts/build_cloud_orchestrator.sh`, not to specify the special way for docker case.

But, I don't want to remove the standalone script, since `go build ./cmd/cloud_orchestrator` itself is relying on current dir. To keep `scripts/docker/run.sh` stable, storing CO binary should be in the known path like root path of git repository, unless users want to specify the path of CO binary. That's the purpose of using standalone script.

docs/cvdr.md
0405ysj: BTW, I think this is out of scope from this change. I think the best way for managing documents would be uploading document for docker case first, and updating it including `HttpBasicAM` or `LocalIM`.

denniscy1993: SGTM!

docs/docker.md
```diff
@@ -0,0 +1,52 @@
...
+
+To build cloud orchestrator, please execute:
+```bash
+git clone https://github.com/google/cloud-android-orchestration.git
+cd cloud-android-orchestration # Root directory of git repository
+scripts/docker/build.sh
```
denniscy1993, line 19: scripts/build_cloud_orchestrator.sh

docs/docker.md
```diff
@@ -0,0 +1,52 @@
...
+```
+
+If build was successful, you would be able to see cloud_orchestrator binary at
+`/path/to/cloud-android-orchestration/cloud_orchestrator`.
```
denniscy1993, line 23: should this be changed to `cloud_orchestrator`?

docs/docker.md
0405ysj: Thanks! I missed that change.

docs/docker.md
0405ysj: Since leaving `cloud_orchestrator` only doesn't seem like the path, so I rewrote the sentence.

denniscy1993 (approved):
LGTM, but I think we also need Sergio's review.

0405ysj:
Though I think he already had a sufficient review, I'll wait for him until tomorrow morning, to submit.

@ser-io Would you have any other opinions?

ikicha (approved):

docs/docker.md
ser-io: > But, I don't want to remove the standalone script, since go build ./cmd/cloud_orchestrator itself is relying on current dir.

I think it's not about current dir in this case, building from the same place ensures Consistency and Reproducibility, this principle is followed by tools like `bazel`, `make` and `go build`, where you need to run the build commands where the project is defined. Default `go build` is expected to be run from where `go.mod` is.

We should follow the same principle for the new `run.sh` script as well, where we should fail if we are not where `go.mod` is, keeping it consistent with build system patterns.

ser-io (changes requested):
This just a reference to the latest comment https://github.com/google/cloud-android-orchestration/pull/308/files#r1598883228 as I couldn't request changes to the PR in that thread.

docs/docker.md
0405ysj: Understood your saying, and dropped my idea to be independent from current dir. So marking this conversation as resolved in advance.

> We should follow the same principle for the new run.sh script as well, where we should fail if we are not where go.mod is, keeping it consistent with build system patterns.

Just out of curiosity, should we really follow the principle? I think what we should follow is build task running well on where `go.mod` is, not throwing an error if we aren't where `go.mod` is.

ser-io: the first step in run commands, i.e `bazel run` or `go run` is always compiling, then the binary is executed. if we follow the same principle, meaning adding `go build ./cmd/cloud_orchestrator` to `scripts/docker/run.sh` then `run.sh` will fail if not run from where the Golang module is defined. 

you could keep the build step in the documentation, however I suggest to compile the binary as part of `run.sh`.

ser-io (approved):

docs/docker.md
0405ysj: > you could keep the build step in the documentation, however I suggest to compile the binary as part of run.sh, this way the user could skip the build part if wanted.

Sounds good, I haven't considered `go run`. Though, I proceeded to merge this PR in advance to publish the document into the public place. I'll update it in the near future.

0405ysj: https://github.com/google/cloud-android-orchestration/pull/311

PR #312, opened 2024-05-17
Elaborate on using SOCKS5 proxy in docs/docker.md

0405ysj:


docs/docker.md
```diff
@@ -43,9 +43,13 @@ TODO(0405ysj): Update this section after renaming the flag name from
 `http_proxy` into `proxy`.
 -->
 If there's a network or security issue for accessing cloud orchestrator
```
ser-io, line 45: > If there's a network or security issue for accessing cloud orchestrator

we could be more specific in the previous sentence: we suggest using a `SOCKS5 proxy` if port 22 is available whereas HO port 2080 might be blocked by firewall. This is the main scenario that comes to mind, feel free to add others if relevant.

ser-io (approved):

docs/docker.md
0405ysj: I thought that wording connection type is rather than numeric port number, I rewrote the sentence for elaborating the network issue.

PR #314, opened 2024-05-21
Rename HTTPBasicAccountManager to UsernameOnlyAccountManager.

denniscy1993:
- Renaming this account manager because it was designed for closed deployed cloud orchestrator. Although we leverage the HTTP basic authentication, only username is used.

- Tests:
  - `cvdr create`
  - `cvdr list`
  - `cvdr delete`

### DIFF
--- a/cmd/cloud_orchestrator/main.go
+++ b/cmd/cloud_orchestrator/main.go
@@ -115,8 +115,8 @@ func LoadAccountManager(config *config.Config) accounts.Manager {
 		am = accounts.NewGAEUsersAccountManager()
 	case accounts.UnixAMType:
 		am = accounts.NewUnixAccountManager()
-	case accounts.HTTPBasicAMType:
-		am = accounts.NewHTTPBasicAccountManager()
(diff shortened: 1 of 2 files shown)


denniscy1993:
Note: I did not change the name of `AuthnOpts.HTTPBasic` at `pkg/client/client.go` because I think this is related to how we build the HTTP requests. We are still leveraging the HTTP basic authentication while building requests even though the name of AM is being changed.

ser-io (changes requested):
let's keep this PR focused on the cloud orchestator (server) code only. 

we can rename the concepts in cvdr (client) in a different PR, although I think it's not necessary in this case.

0405ysj (approved):
LGTM, but please resolve Sergio's comment in advance.

denniscy1993:
> let's keep this PR focused on the cloud orchestator (server) code only.
> 
> we can rename the concepts in cvdr (client) in a different PR, although I think it's not necessary in this case.

Done, and agree that we don't need to rename the client.

ser-io (approved):

PR #315, opened 2024-05-24
Multi service

ser-io:


pkg/cli/config.go
```diff
@@ -53,16 +53,33 @@ type HTTPBasicAuthnConfig struct {
...
 	HTTPProxy                 string       `json:"http_proxy,omitempty"`
-	ConnectionControlDir      string       `json:"connection_control_dir,omitempty"`
-	KeepLogFilesDays          int          `json:"keep_log_files_days,omitempty"`
 	BuildAPICredentialsSource string       `json:"build_api_credentials_source,omitempty"`
-	Host                      HostConfig   `json:"host,omitempty"`
+	Host                      *HostConfig  `json:"host,omitempty"`
 	Authn                     *AuthnConfig `json:"authn,omitempty"`
 }
 
+func (c *Config) Service() *Service {
```
jemoreira, line 74: This looks like it should be named `DefaultService`.

jemoreira (approved):

pkg/cli/config.go
ser-io: yes indeed. done.

pkg/cli/cli.go
```diff
@@ -446,10 +446,9 @@ func hostCommand(opts *subCommandOpts) *cobra.Command {
 		},
 	}
 	create.Flags().StringVar(&createFlags.GCP.MachineType, gcpMachineTypeFlag,
-		opts.InitialConfig.Host.GCP.MachineType, gcpMachineTypeFlagDesc)
+		opts.InitialConfig.DefaultService().Host.GCP.MachineType, gcpMachineTypeFlagDesc)
 	create.Flags().StringVar(&createFlags.GCP.MinCPUPlatform, gcpMinCPUPlatformFlag,
-		opts.InitialConfig.Host.GCP.MinCPUPlatform, gcpMinCPUPlatformFlagDesc)
-	create.Flags().StringArrayVar(&acceleratorFlagValues, acceleratorFlag, nil, acceleratorFlagDesc)
+		opts.InitialConfig.DefaultService().Host.GCP.MinCPUPlatform, gcpMinCPUPlatformFlagDesc)
```
0405ysj, line 451: Why do we set the default value as configuration of default service? In case of given service is not a default one, shouldn't other flags be depending on what the service is?

PR #317, opened 2024-05-30
Redirect to the `/username` endpoint on authentication failure.

denniscy1993:
Only `UsernameOnlyAccountManager` needs this redirection, so we extend the `UserFromRequest` interface to delegate this logic to account manager implementations.


pkg/app/accounts/usernameonly.go
```diff
@@ -38,8 +38,12 @@ func NewUsernameOnlyAccountManager() *UsernameOnlyAccountManager {
...
 }
 
-func (m *UsernameOnlyAccountManager) UserFromRequest(r *http.Request) (User, error) {
-	return userFromRequest(r)
+func (m *UsernameOnlyAccountManager) UserFromRequest(w http.ResponseWriter, r *http.Request) (User, error) {
+	user, err := userFromRequest(r)
+	if user == nil {
+		http.Redirect(w, r, "/username", http.StatusSeeOther)
+	}
+	return user, err
```
0405ysj, line 46: Out of curiosity, is `err` still valid though redirection to `/username` brings the cookie, in case of CO WebUI?

pkg/app/accounts/usernameonly.go
```diff
@@ -38,8 +38,12 @@ func NewUsernameOnlyAccountManager() *UsernameOnlyAccountManager {
 	return &UsernameOnlyAccountManager{}
 }
 
-func (m *UsernameOnlyAccountManager) UserFromRequest(r *http.Request) (User, error) {
-	return userFromRequest(r)
+func (m *UsernameOnlyAccountManager) UserFromRequest(w http.ResponseWriter, r *http.Request) (User, error) {
```
ser-io, line 41: let's avoid modifying the `Manager` interface in this case.  We can add the redirection logic directly in the handler here: https://github.com/google/cloud-android-orchestration/blob/main/pkg/app/app.go#L517, if `user == nil` return a typed error, catch the error in the handler in redirect from there. here's a good error about errors and errors handling in golang: https://go.dev/blog/error-handling-and-go

ser-io (changes requested):

pkg/app/accounts/usernameonly.go
denniscy1993: I think you are correct. `err` should be no longer valid after successfully logging in by redirecting to `/username`. I removed the error code response in my new patch set.

pkg/app/accounts/usernameonly.go
denniscy1993: Thanks for sharing the Golang good practices, super helpful!

I'm wondering if it would be better to delegate the trigger of redirection to `UsernameOnlyAccountManager`, because that only this account manager type needs that. So I would like to return this new error type at `UsernameOnlyAccountManager` instead of when `app.go` receives a `nil` user. What do you think?

pkg/app/app.go
```diff
@@ -536,6 +536,10 @@ func (h HTTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		log.Println("Error: ", err)
 		var e *apperr.AppError
 		if errors.As(err, &e) {
+			if e.Error() == accounts.MissingUsername {
+				http.Redirect(w, r, "/username", http.StatusSeeOther)
```
ikicha, line 540: I am not sure about what de-factor is, but it looks like 302(StatusFound) is usually used for login redirect

ikicha, line 540: And can we append a kind of 'callback' url as suffix to move back to current page after login?

pkg/app/accounts/usernameonly.go
0405ysj: Resolving this. I think this thread is now outdated.

pkg/app/app.go
```diff
@@ -536,6 +536,10 @@ func (h HTTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		log.Println("Error: ", err)
 		var e *apperr.AppError
 		if errors.As(err, &e) {
+			if e.Error() == accounts.MissingUsername {
```
ser-io, line 539: can this be rather added in the `Authenticate` method handler above? there we can make sure this condition is only added if using the `UsernameOnly` account manager.

pkg/app/accounts/usernameonly.go
ser-io: yes, you got it. the new error should be returned from here in `UsernameOnlyAccountManager.UserFromRequest`.

pkg/app/app.go
denniscy1993: We can. How about redirecting back to the original page when `HandleUsernameOnlyLogging` gets a valid username from the POST method?

pkg/app/app.go
denniscy1993: I moved this redirection to the `Authenticate` method. Also, I think if we put the redirection logic there, we can redirect when receiving a `nil` user from `UsernameOnlyAccountManager`. It means that we don't need to define a new error type. WDYT?

ser-io (approved):

pkg/app/app.go
ser-io: LGTM

0405ysj (approved):

PR #319, opened 2024-06-03
Accelerators

ser-io:


pkg/cli/host.go
```diff
@@ -28,6 +30,8 @@ type CreateHostOpts struct {
 type CreateGCPHostOpts struct {
 	MachineType    string
 	MinCPUPlatform string
+	// List of comma separated `key=value` pairs, i.e  `type=nvidia-tesla-p100,count=1`.
+	AcceleratorConfigs []string
```
Databean, line 34: Can this be a string->string map?

https://google.aip.dev/apps/2717 suggests a string->string map for storing generic data, and [go/apidosdonts#dont-encode-data-in-a-string](http://go/apidosdonts#dont-encode-data-in-a-string) writes

> Don’t Encode Data in a String That You Expect a Client to Construct or Parse

Databean (approved):

pkg/cli/host.go
ser-io: Done. This struct is not meant to be serialized, however I moved the parsing logic to an early stage allowing us to define a more specific structure. Thanks.

PR #322, opened 2024-06-05
Update documents for UsernameOnlyAccountManager.

denniscy1993:


docs/docker.md
```diff
@@ -39,5 +35,14 @@ Besides, currently `cvdr` supports using SOCKS5 proxy with the flag like
...
 
-Please follow [cvdr.md](cvdr.md). The URL of running cloud orchestrator should
-be `http://${CLOUD_ORCHESTRATOR_IP_ADDRESS}:8080`.
+Please use the configuration file `scripts/docker/cvdr.toml` and follow the
+steps at [cvdr.md](cvdr.md). The URL of running cloud orchestrator should
+be `http://${CLOUD_ORCHESTRATOR_IP_ADDRESS}:8080`. For example:
+```bash
+CVDR_USER_CONFIG_PATH=./scripts/docker/cvdr.toml \
+./cvdr \
+--service_url=http://${CLOUD_ORCHESTRATOR_IP_ADDRESS}:8080 \
```
ser-io, line 44: if using a config file `scripts/docker/cvdr.toml` already then we can skip declaring the `service_url`, `proxy` and `zone` flags on every execution, make these values part of the configuration file as well.

ser-io (changes requested):

docs/docker.md
denniscy1993: Done, thanks!

ser-io (approved):

docs/docker.md
```diff
@@ -39,5 +35,8 @@ Besides, currently `cvdr` supports using SOCKS5 proxy with the flag like
 
 ## Use cloud orchestrator by cvdr
 
-Please follow [cvdr.md](cvdr.md). The URL of running cloud orchestrator should
-be `http://${CLOUD_ORCHESTRATOR_IP_ADDRESS}:8080`.
+Please use the configuration file `scripts/docker/cvdr.toml` and follow the
+steps at [cvdr.md](cvdr.md). For example:
+```bash
+CVDR_USER_CONFIG_PATH=./scripts/docker/cvdr.toml ./cvdr host create
```
0405ysj, line 41: Sorry for late reply, could we mention how to use `cvdr` into `docs/cvdr.md`, not `docs/docker.md`? Also, could you remove all `--service_url`s in `docs/cvdr.md` and replace with `CVDR_USER_CONFIG_PATH`?

The main reason for this suggestion is `docs/cvdr.md` suggests to find `SERVICE_URL` value, but nowhere(docs/cvdr.md`, `docs/cloud_orchestrator.md`, `docs/docker.md`) doesn't explain what's `SERVICE_URL`.

0405ysj (changes requested):

docs/docker.md
denniscy1993: I think we would still need a `cvdr` section at `docs/docker.md`. This is because we have a specific `./scripts/docker/cvdr.toml` for the docker use case.

And regarding replacing `--service_url`s in `docs/cvdr.md` with `CVDR_USER_CONFIG_PATH`, which `cvdr.toml` would you suggest us to use? Should we declare a new one or should we just refer to `build/package/host/etc/cvdr.toml`?

0405ysj: > I think we would still need a cvdr section at docs/docker.md.

Agree, I thought there still needs `cvdr` section, but thought no needs to put extra information beyond explaining the path of `CVDR_USER_CONFIG_PATH`. Then, `docs/cvdr.md` should take the responsibility explaining about how to use `CVDR_USER_CONFIG_PATH`.

> And regarding replacing --service_urls in docs/cvdr.md with CVDR_USER_CONFIG_PATH, which cvdr.toml would you suggest us to use?

As I wrote about SERVICE_URL before, I think `docs/cvdr.md` needs a section providing the redirection link towards like `docs/docker.md`. For instance, setting the title of section as `Prepare cvdr.toml`, and giving a suggestion with that link.

denniscy1993: SGTM! Done.

0405ysj (approved):

docs/docker.md
0405ysj: Thanks :)

ikicha (approved):

denniscy1993:
https://github.com/google/cloud-android-orchestration/pull/315 this PR modified the structure of `cvdr.toml`, so I also changed `scripts/docker/cvdr.toml` in this PR.

PR #323, opened 2024-06-05
Provide more info when json decoding fails.

ser-io:


pkg/client/httputils.go
```diff
@@ -165,7 +165,25 @@ func (rb *HTTPRequestBuilder) JSONResDoWithRetries(ret any, retryOpts RetryOptio
...
+	}
+	decoder := json.NewDecoder(bytes.NewReader(b))
+	if res.StatusCode >= 200 && res.StatusCode <= 299 {
+		if ret == nil {
+			return nil
+		}
+		if err := decoder.Decode(ret); err != nil {
+			return fmt.Errorf("failed decoding successful response(%d), body: %s, error: %w", res.StatusCode, string(b), err)
+		}
+		return nil
```
Databean, line 180: Is it important that a successful response decodes to valid json when the response is dropped anyway?

Databean (approved):

pkg/client/httputils.go
Databean: My mistake, I didn't realize the caller of this function similarly has access to `ret`.

ser-io: no problem! thanks for reviewing.

PR #325, opened 2024-06-06
Use a max wait threshold in retry logic rather than number of retries.

ser-io:


pkg/client/httputils.go
```diff
@@ -205,7 +207,14 @@ func (rb *HTTPRequestBuilder) doWithRetries(retryOpts RetryOptions) (*http.Respo
...
 		return nil, fmt.Errorf("error sending request: %w", err)
 	}
-	for i := uint(0); i < retryOpts.NumRetries && isIn(res.StatusCode, retryOpts.StatusCodes); i++ {
+	maxWaitElapsed := atomic.Bool{}
+	maxWaitElapsed.Store(false)
+	go func() {
+		time.Sleep(retryOpts.MaxWait)
+		maxWaitElapsed.Store(true)
+	}()
+	elpased := maxWaitElapsed.Load()
```
jemoreira, line 216: An extra goroutine shouldn't be necessary for this, here is a simpler alternative:

```
timeoutTime := time.Now() + timeoutDelay
success := false
for ; time.Now() < timeoutTime; {
  if err := doTheWork(); err != nil {
    // maybe log
    continue
  }
  success = true
}

if !success {
  ...
}

```

ser-io: ufff, true, my mistake, thanks for spotting this. done.

jemoreira (approved):

PR #334, opened 2024-07-17
Fix double closure of sockets for websocket_agent

k311093:
At websocket_agent, wsWrapper.Close() is executed twice, at deferred function and inside go routine. Call wsWrapper.Close() only at go routine that writes data to the wsWrapper.

Also move tcpConn.Close() after io.Copy(...) to avoid confusion.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -1272,7 +1272,6 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
 		c.PrintErrf("Failed to accept ADB socket: %v", err)
 		return err
 	}
-	defer tcpConn.Close()
 
 	// Connect to ADB proxy WebSocket
 	//   wss://127.0.0.1:1443/devices/cvd-1/adb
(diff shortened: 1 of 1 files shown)


pkg/cli/cli.go
```diff
@@ -1321,12 +1320,12 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
 		pos:    0,
 		buf:    nil,
 	}
-	defer wsWrapper.Close()
 	go func() {
 		io.Copy(wsWrapper, tcpConn)
 		wsWrapper.Close()
 	}()
 	io.Copy(tcpConn, wsWrapper)
+	tcpConn.Close()
```
jemoreira, line 1328: This one should stay as it was, with the `defer` in line 1275. It has the same effect of closing at the end of the function, but with the added benefit of closing the socket even with an early return in case of errors.

I wrongly suggested before that this one needed fixing because I had the order of parameters to `io.Copy` inverted. Sorry about that.

jemoreira (changes requested):

PR #335, opened 2024-07-22
Parse system and user configuration files independently.

ser-io:
- Parsing a combined blob is not correct as top level fields needs to be listed at the beginning for a correct parsing.

### DIFF
--- a/cmd/cvdr/main.go
+++ b/cmd/cvdr/main.go
@@ -35,10 +35,9 @@ const (
 
 func loadInitialConfig() (*cli.Config, error) {
 	config := cli.BaseConfig()
-	srcs := []string{}
+	sysConfigSrc, userConfigSrc := "", ""
 	if path, ok := os.LookupEnv(envVarSystemConfigPath); ok {
-		path = cli.ExpandPath(path)
-		srcs = append(srcs, path)
+		sysConfigSrc = cli.ExpandPath(path)
 	}
 	if path, ok := os.LookupEnv(envVarUserConfigPath); ok {
 		path = cli.ExpandPath(path)
@@ -62,13 +61,12 @@ func loadInitialConfig() (*cli.Config, error) {
 			}
 			statErr = nil
 		}
-		if statErr == nil {
-			srcs = append(srcs, path)
-		} else {
+		if statErr != nil {
 			return nil, fmt.Errorf("invalid user config file path: %w", statErr)
 		}
+		userConfigSrc = path
 	}
-	if err := cli.LoadConfig(srcs, config); err != nil {
+	if err := cli.LoadConfig(sysConfigSrc, userConfigSrc, config); err != nil {
 		return nil, err
 	}
 	return config, nil
--- a/pkg/cli/config.go
+++ b/pkg/cli/config.go
@@ -95,20 +95,25 @@ func BaseConfig() *Config {
 	}
 }
 
-func LoadConfig(srcs []string, c *Config) error {
-	combined := []byte{}
-	for _, path := range srcs {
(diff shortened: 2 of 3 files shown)


cmd/cvdr/main.go
```diff
@@ -63,12 +62,12 @@ func loadInitialConfig() (*cli.Config, error) {
 			statErr = nil
 		}
 		if statErr == nil {
-			srcs = append(srcs, path)
+			userConfigSrc = path
 		} else {
 			return nil, fmt.Errorf("invalid user config file path: %w", statErr)
 		}
```
jemoreira, line 68: nit: Follow this pattern
```
if statErr != nil {
    return ...
}
userConfigSrc = path
```

pkg/cli/config.go
```diff
@@ -95,20 +95,44 @@ func BaseConfig() *Config {
...
 	}
-	decoder := toml.NewDecoder(bytes.NewReader(combined))
-	// Fail if there is some unknown configuration. This is better than silently
-	// ignoring a (perhaps misspelled) config entry.
-	decoder.Strict(true)
-	return decoder.Decode(c)
+	tmpSysConfig := &Config{}
+	if err := loadFn(sysSrc, tmpSysConfig); err != nil {
+		return fmt.Errorf("error loading system configuration: %w", err)
+	}
```
jemoreira, line 126: Instead of loading system configs twice, would it be possible to store the `Services` dictionary from the first time it was loaded?

pkg/cli/config.go
```diff
@@ -95,20 +95,44 @@ func BaseConfig() *Config {
 	}
 }
 
-func LoadConfig(srcs []string, c *Config) error {
-	combined := []byte{}
-	for _, path := range srcs {
-		b, err := os.ReadFile(path)
+func LoadConfig(sysSrc, userSrc string, c *Config) error {
+	loadFn := func(src string, out *Config) error {
```
jemoreira, line 99: nit: I'd prefer this function to be outside `LoadConfig`. Maybe renaming `LoadConfig` as `LoadConfigs` and naming this one `LoadConfig` or `ExtendConfigFromFile`

ser-io: Done, moved it to package private function: `func loadConfig(src string, out *Config) error`.

pkg/cli/config.go
ser-io: Done.

cmd/cvdr/main.go
ser-io: Good catch! Done.

pkg/cli/config.go
```diff
@@ -95,20 +95,24 @@ func BaseConfig() *Config {
...
 	}
-	decoder := toml.NewDecoder(bytes.NewReader(combined))
-	// Fail if there is some unknown configuration. This is better than silently
-	// ignoring a (perhaps misspelled) config entry.
-	decoder.Strict(true)
-	return decoder.Decode(c)
+	if userSrc == "" {
+		return nil
+	}
+	sysSrvcs := c.Services
```
jemoreira, line 106: While probably not a problem in practice, it's better to make a copy of the elements or set `c.Services` to `nil` after this assignment. `c` is passed by reference and is expected to be modified by the json package, which may or may not override this property. AFAIK it's not guaranteed that it will not reuse the underlying array in `c.Services`, of which `sysSrvcs` is also a reference of (a slice).

ser-io: True, I loaded the system config a second time earlier to have a deep copy of the system services and avoid using the `reflect` package. Went with `c.Services = nil` in this case.

jemoreira (approved):

PR #340, opened 2024-07-30
Smalls improvements in flags structures.

ser-io:


jemoreira (approved):
A more descriptive name for the PR would be much appriaciated.

ser-io:
> A more descriptive name for the PR would be much appriaciated.

Agree.

PR #344, opened 2024-08-07
Support service account credential authentication flow

chyijia:
- Enable getting service account credential from SecretManager
- Add fields in Oauth2.Helper for authentication flow
- Implement JWT Authentication flow

### DIFF
--- a/cmd/cloud_orchestrator/main.go
+++ b/cmd/cloud_orchestrator/main.go
@@ -98,14 +98,30 @@ func LoadSecretManager(config *config.Config) secrets.SecretManager {
 }
 
 func LoadOAuth2Config(config *config.Config, sm secrets.SecretManager) *appOAuth2.Helper {
-	var oauth2Helper *appOAuth2.Helper
+	var credential []byte
+	switch config.SecretManager.Type {
+	case secrets.UnixSMType:
+		secretFilePath := config.SecretManager.UNIX.SecretFilePath
+		var err error
+		credential, err = sm.(*secrets.FromFileSecretManager).GetServiceAccountCredential(secretFilePath)
+		if err != nil {
+			log.Fatal(err)
+		}
+	default:
+		credential = []byte{}
+	}
+
 	switch config.AccountManager.OAuth2.Provider {
 	case appOAuth2.GoogleOAuth2Provider:
-		oauth2Helper = appOAuth2.NewGoogleOAuth2Helper(config.AccountManager.OAuth2.RedirectURL, sm)
+		oauth2Helper, err := appOAuth2.NewGoogleOAuth2Helper(config.AccountManager.OAuth2.RedirectURL, sm, credential)
+		if err != nil {
+			log.Fatal(err)
+		}
+		return oauth2Helper
 	default:
 		log.Fatal("Unknown oauth2 provider: ", config.AccountManager.OAuth2.Provider)
 	}
-	return oauth2Helper
+	return nil
 }
 
 func LoadAccountManager(config *config.Config) accounts.Manager {
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -251,19 +251,34 @@ func (c *App) waitOperation(w http.ResponseWriter, r *http.Request, user account
 }
 
 func (c *App) AuthHandler(w http.ResponseWriter, r *http.Request) error {
-	state := randomHexString()
-	s := session.Session{
-		OAuth2State: state,
+	switch c.oauth2Helper.AuthType {
+	case appOAuth2.JWTAuthType:
+		tk, err := c.oauth2Helper.JwtConfig.TokenSource(context.Background()).Token()
+		if err != nil {
+			return err
+		}
+		user, err := c.accountManager.UserFromRequest(r)
+		if err != nil {
+			return err
+		}
+		if err := c.storeUserCredentials(user, tk); err != nil {
+			return err
+		}
+		return nil
+	case appOAuth2.OAuthClientType:
+		state := randomHexString()
+		s := session.Session{
+			OAuth2State: state,
+		}
+		if err := c.setOrUpdateSession(w, &s); err != nil {
+			return err
+		}
+		authURL := c.oauth2Helper.OAuthConfig.AuthCodeURL(state, oauth2.AccessTypeOffline)
+		http.Redirect(w, r, authURL, http.StatusSeeOther)
+		return nil
+	default:
+		return errors.New("Invalid auth type")
 	}
-	if err := c.setOrUpdateSession(w, &s); err != nil {
-		return err
-	}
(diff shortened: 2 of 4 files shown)


ser-io (changes requested):
Hi @chyijia, thanks for your contribution. For an easier review we can split this commit into two smaller ones, one for accepting service account credentials and the other for the JWT flow. You can read more about the benefits of smaller units of change at [go/small-cls](https://goto.google.com/small-cls).

pkg/app/secrets/local.go
```diff
@@ -52,3 +52,26 @@ func (sm *FromFileSecretManager) OAuth2ClientID() string {
...
 }
+
+type CredentialType struct {
+	Type string `json:"type"`
+}
+
+// Check if the credential type is service accout,
+// if so: return credential content
+// else: return nil
+func GetServiceAccountCredential(path string) ([]byte, error) {
```
0405ysj, line 63: I think the feature getting service account credential is bound into `UnixSMType`. What about doing this logic in `NewFromFileSecretManager`, storing credential in the struct of `FromFileSecretManager`, and retrieving the credential by calling a new method of `FromFileSecretManager` like `OAuth2ClientID()` or `OAuth2ClientSecret()` does?

pkg/app/app.go
```diff
@@ -258,7 +258,7 @@ func (c *App) AuthHandler(w http.ResponseWriter, r *http.Request) error {
 	if err := c.setOrUpdateSession(w, &s); err != nil {
 		return err
 	}
-	authURL := c.oauth2Helper.AuthCodeURL(state, oauth2.AccessTypeOffline)
+	authURL := c.oauth2Helper.OAuthConfig.AuthCodeURL(state, oauth2.AccessTypeOffline)
```
0405ysj: I think we should care when `helper.OAuthConfig` is nil. Unless we have a check on OAuthConfig, we would face at unexpected behavior.

pkg/app/app.go
```diff
@@ -286,7 +286,7 @@ func (c *App) OAuth2Callback(w http.ResponseWriter, r *http.Request) error {
 	if err != nil {
 		return err
 	}
-	tk, err := c.oauth2Helper.Exchange(context.Background(), authCode)
+	tk, err := c.oauth2Helper.OAuthConfig.Exchange(context.Background(), authCode)
```
0405ysj, line 307: Ditto, I think we should care when `helper.OAuthConfig` is nil. Unless we have a check on OAuthConfig, we would face at unexpected behavior.

pkg/app/app.go
```diff
@@ -498,7 +498,7 @@ func (c *App) fetchUserCredentials(user accounts.User) (*oauth2.Token, error) {
 	}
 	if !tk.Valid() {
 		// Refresh the token and store it in the db.
-		tks := c.oauth2Helper.TokenSource(context.TODO(), tk)
+		tks := c.oauth2Helper.OAuthConfig.TokenSource(context.TODO(), tk)
```
0405ysj, line 519: Ditto, I think we should care when `helper.OAuthConfig` is nil. Unless we have a check on OAuthConfig, we would face at unexpected behavior.

pkg/app/oauth2/oauth2.go
```diff
@@ -35,15 +36,37 @@ const (
 	GoogleOAuth2Provider = "Google"
 )
 
+const (
+	JWTAuthType     = "jwt"
+	OAuthClientType = "oauth"
```
0405ysj, line 41: What about introduce a new OAuthType like other managers(e.g. InstanceManager, AccountManager, SecretManager) do?

0405ysj (changes requested):
Thanks for your contribution :)

pkg/app/oauth2/oauth2.go
denniscy1993: Good suggestion, this should make the code more readable, but an open question here is how should we classify these auth types?

What @chyijia tries to add here is the JWT authentication flow, which does not belong to the OAuth protocol if I understand correctly. Also, CO currently uses `Provider` to decide which `oauth2Helper` should be created (https://github.com/google/cloud-android-orchestration/blob/main/cmd/cloud_orchestrator/main.go#L100). However, in our case, I think this should be based on which authentication protocol is using. So how about defining two different AuthHelpers, one is `GoogleOAuth2Helper` and another is `JWTHelper`?

We may need @ser-io's suggestions about this small refactor.

pkg/app/secrets/local.go
denniscy1993: Sounds good, but I think it would be better to define a standalone function at `FromFileSecretManager` rather than extending the `SecretManager` interfaces. The reason is that this service account credentials should not be used for other secret manger types.

pkg/app/oauth2/oauth2.go
denniscy1993: To make it clear, I mean something like:
```
switch config.AccountManager.Auth.Type {
case JWT:
	helper = NewJWTHelper(...)
case OAuth2:
	switch config.AccountManager.Auth.OAuth2.Provider {
	case appOAuth2.GoogleOAuth2Provider:
		helper = NewGoogleOAuth2Helper(...)
	default:
		...
	}
default:
	...
}
WDYT?

Comment (0405ysj) [reply pkg/app/oauth2/oauth2.go:41]: Ah, I got a misunderstanding about JWT. It wasn't a part of OAuth protocol. Sorry for giving confusion.

I think having 2 different AuthHelper is fine. And not sure about listing type inside of `AccountManager`, but no strong opinion from my side. Your suggestion SGTM at least, but prefer to listen @ser-io. He might have a different view.

File: cmd/cloud_orchestrator/main.go [hunk 6]
```diff
@@ -99,9 +99,23 @@ func LoadSecretManager(config *config.Config) secrets.SecretManager {
 
 func LoadOAuth2Config(config *config.Config, sm secrets.SecretManager) *appOAuth2.Helper {
 	var oauth2Helper *appOAuth2.Helper
+	var credential []byte
+	if config.SecretManager.Type == secrets.UnixSMType {
+		secretFilePath := config.SecretManager.UNIX.SecretFilePath
+		var err error
+		credential, err = secrets.GetServiceAccountCredential(secretFilePath)
```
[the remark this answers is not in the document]
chyijia: If you're referring to the `credential, err :=` part on L105, I declared it outside the if block because the block might not always be executed, but the credential variable could potentially be used on L113.

 Is my understanding correct?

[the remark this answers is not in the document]
0405ysj: Ah, I missed what you said during writing review comments. You're right, never mind.

chyijia:
Close this PR because there is another solution. #348

PR #346, opened 2024-08-12
Enable UsernameOnlyAccountManager to support OAuth

chyijia:
Fixes email verification failure in UsernameOnlyAccountManager during OAuth by supporting the Email field.

- For web use case:
    - Add email field in login page
    - Store the entered email in the cookie for further use
  
- For CLI use case:
    - enable setting `UserEmail = <email>` in `cvdr.toml` to configure the email in `UsernameOnlyUser`. For example:
```
Authn = {
  HTTPBasicAuthn = {
    UsernameSrc = "unix",
    UserEmail = <Account used for OAuth>
  }
}
```

### DIFF
--- a/pkg/app/accounts/usernameonly.go
+++ b/pkg/app/accounts/usernameonly.go
@@ -26,6 +26,8 @@ const (
 
 	unameCookie    = "accountUsername"
 	userMailCookie = "accountEmail"
+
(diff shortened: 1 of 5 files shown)


chyijia:
We've decided to obtain and manage access tokens using cvdr (#348). We don't need this for now.

PR #348, opened 2024-08-20
Support JWT auth flow that access tokens are managed by cvdr

chyijia:
For more information, please refer to b/353884144#comment13

### DIFF
--- a/pkg/cli/cvd.go
+++ b/pkg/cli/cvd.go
@@ -16,6 +16,7 @@ package cli
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -26,6 +27,7 @@ import (
 	"text/template"
 
 	"github.com/google/cloud-android-orchestration/pkg/client"
+	"golang.org/x/oauth2/google"
 
 	hoapi "github.com/google/android-cuttlefish/frontend/src/host_orchestrator/api/v1"
 	"github.com/hashicorp/go-multierror"
@@ -356,8 +358,24 @@ func credentialsFactoryFromSource(source string) (CredentialsFactory, error) {
 	case InjectedCredentialsSource:
 		return func() string { return client.InjectedCredentials }, nil
 	default:
-		return nil, fmt.Errorf("unknown credentials source: %s", source)
+		return jwtAccessToken(source)
 	}
+}
+
+func jwtAccessToken(filepath string) (func() string, error) {
+	content, err := os.ReadFile(filepath)
+	if err != nil {
+		return nil, fmt.Errorf("cannot read content from credential filepath %s: %w", filepath, err)
+	}
+	jwtConfig, err := google.JWTConfigFromJSON(content, "https://www.googleapis.com/auth/androidbuild.internal")
+	if err != nil {
+		return nil, err
+	}
+	tk, err := jwtConfig.TokenSource(context.Background()).Token()
+	if err != nil {
+		return nil, err
+	}
+	return func() string { return tk.AccessToken }, nil
 }
 
 type cvdListResult struct {
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -67,12 +67,13 @@ type HTTPBasic struct {
 }
 
 type ServiceOptions struct {
-	RootEndpoint   string
-	ProxyURL       string
-	DumpOut        io.Writer
-	ErrOut         io.Writer
-	ChunkSizeBytes int64
-	Authn          *AuthnOpts
+	RootEndpoint        string
+	ProxyURL            string
+	DumpOut             io.Writer
+	ErrOut              io.Writer
+	ChunkSizeBytes      int64
+	Authn               *AuthnOpts
+	InjectBuildAPICreds bool
 }
 
 type Service interface {
@@ -189,7 +190,10 @@ func (s *serviceImpl) HostService(host string) HostOrchestratorService {
 	hs := &HostOrchestratorServiceImpl{
 		HTTPHelper: s.httpHelper,
 		// Make the cloud orchestrator inject the credentials instead
-		BuildAPICredentialsHeader: headerNameCOInjectBuildAPICreds,
+		BuildAPICredentialsHeader: DefaultHostOrchestratorCredentialsHeader,
+	}
+	if s.InjectBuildAPICreds {
+		hs.BuildAPICredentialsHeader = headerNameCOInjectBuildAPICreds
 	}
 	hs.HTTPHelper.RootEndpoint = s.httpHelper.RootEndpoint + "/hosts/" + host
 	return hs
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -1593,6 +1593,7 @@ func buildServiceBuilder(builder client.ServiceBuilder, config *Config) serviceB
 				}
 			}
 		}
+		opts.InjectBuildAPICreds = srvConfig.BuildAPICredentialsSource == "injected"
 		return builder(opts)
 	}
 }
--- a/pkg/client/host_orchestrator_client.go
+++ b/pkg/client/host_orchestrator_client.go
@@ -67,7 +67,7 @@ type HostOrchestratorService interface {
 	CreateBugreport(group string, dst io.Writer) error
 }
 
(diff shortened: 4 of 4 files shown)


pkg/cli/cvd.go
```diff
@@ -101,6 +106,7 @@ type CreateCVDOpts struct {
 	// If true, perform the ADB connection automatically.
 	AutoConnect               bool
 	BuildAPICredentialsSource string
+	CredentialFilePath        string
```
denniscy1993, line 109: Would it be better to combine these two strings in a struct? Because I think they would be used at the same time. For example:
```
type BuildAPICredentialsOpts struct {
    Source string
    FilePath string
}
```

pkg/cli/cvd.go
```diff
@@ -308,15 +318,65 @@ func (c *cvdCreator) createCVDFromLocalSrcs() ([]*hoapi.CVD, error) {
...
+
+// TODO: OAuth Config for end user
+
+func parseAuthConfig(credential []byte) (OAuthConfig, error) {
+	jwtConfig, err := NewJwtConfig(credential)
+	if err == nil {
+		return jwtConfig, nil
+	}
+	return nil, err
+}
```
0405ysj, line 364: Do we need this function? IIUC, this function seems for obtaining `OAuthConfig` type object from `JwtConfig` type object. Couldn't we make it simpler without introducing a new function?

pkg/cli/cvd.go
```diff
@@ -308,15 +318,65 @@ func (c *cvdCreator) createCVDFromLocalSrcs() ([]*hoapi.CVD, error) {
...
+		return auth(opts.FilePath)
 	default:
-		return nil, fmt.Errorf("unknown credentials source: %s", source)
+		return nil, fmt.Errorf("unknown credentials source: %s", opts.Source)
+	}
+}
+
+type OAuthConfig interface {
+	auth() (*oauth2.Token, error)
+}
```
0405ysj, line 336: Could you elaborate the purpose of introducing `OAuthConfig`?

pkg/cli/cvd.go
```diff
@@ -308,15 +318,65 @@ func (c *cvdCreator) createCVDFromLocalSrcs() ([]*hoapi.CVD, error) {
...
+	}
+	return nil, err
+}
+
+func auth(filepath string) (func() string, error) {
+	content, err := os.ReadFile(filepath)
+	if err != nil {
+		return nil, fmt.Errorf("cannot read content from credential filepath %s: %w", filepath, err)
+	}
+	config, err := parseAuthConfig(content)
```
0405ysj, line 371: Without using `parseAuthConfig`, perhaps we can change the code like below? Omitted error handling code in below example just for convenience.
```
config, err := NewJwtConfig(content)
tk, err := (OAuthConfig(config)).auth()
```

0405ysj (changes requested):

pkg/cli/cvd.go
chyijia: The reason why I use a new function is because I'd like to implement 3LO auth flow for end-users use cases probably in next PR, which requires a different type of config. 

To answer the three questions together, the final code would look like this:
```
type OAuthConfig struct {*oauth2.Config}

func (c OAuthConfig) auth(*oauth2.Token, error) {
    // do 3LO auth flow
}

func parseAuthConfig(credential []byte) (OAuthConfig, error) {
	jwtConfig, err := NewJwtConfig(credential)
	if err == nil {
		return jwtConfig, nil
	}
        // if not jwtConfig, try to parse oauth2 config and return
        // else return parse error
}
```
So that the `func auth()` can do the authentication without considering which flow to use.

Does that make sense?

0405ysj: I think it makes sense, but I prefer to change `OAuthConfig` interface type into the different name then. I think your purpose is having a interface type for common uses, and 2 struct types(`OAuthConfig` and `JwtConfig`) implements `auth()` for satisfying interface type. What about `AuthConfig` instead?

BTW, why don't we make a fallback mechanism on `func auth(filepath string)`? For instance, executing `JWTConfigFromJSON` and `TokenSource` in a single function called by `func auth(filepath string)`?

0405ysj (approved):
LGTM. Perhaps we need @ser-io's review unless waiting for a long time.

ser-io:
> LGTM. Perhaps we need @ser-io's review unless waiting for a long time.

@chyijia thanks for resolving existing comments, I'll be reviewing the PR tomorrow.

denniscy1993 (approved):

pkg/cli/cvd.go
```diff
@@ -102,8 +111,8 @@ type CreateCVDOpts struct {
 	// Example: https://cs.android.com/android/platform/superproject/main/+/main:device/google/cuttlefish/host/cvd_test_configs/main_phone-main_watch.json;drc=b2e8f4f014abb7f9cb56c0ae199334aacb04542d
 	EnvConfig map[string]interface{}
 	// If true, perform the ADB connection automatically.
-	AutoConnect               bool
-	BuildAPICredentialsSource string
```
ser-io, line 106: let's continue using `BuildAPICredentialsSource` only using a valid filepath value to it. For example, the config parameter could have  special values like: `none`, `injected` and a file path value like: `/etc/cvdr/token.json`. Remember to update the commit message.

pkg/cli/cvd.go
```diff
@@ -349,15 +358,43 @@ func (c *cvdCreator) createCVDFromLocalSrcs() ([]*hoapi.CVD, error) {
...
 		return func() string { return client.InjectedCredentials }, nil
+	case FileCredentialSource:
+		return auth(opts.FilePath)
 	default:
-		return nil, fmt.Errorf("unknown credentials source: %s", source)
+		return nil, fmt.Errorf("unknown credentials source: %s", opts.Source)
+	}
+}
+
+func jwtAuth(credential []byte) (*oauth2.Token, error) {
```
ser-io, line 374: let's have only one function here for now named: `jwtAccessToken(filepath string)` rather than having `auth` and `jwtAuth`. we can do the split later when we have the implementation for OAuth if necessary, but let's avoid early refactoring.

Suggested name `jwtAccessToken` to be specific, in this context, terms like `auth` and `jwtAuth` are too generic, we are working to get an access token.

ser-io (changes requested):
Let's update the commit message.

For context: when we mention `authentication in cvdr` alone, the first though is authenticating `cvdr` against the Cloud Orchestrator, which is the service `cvdr` connects to. 

In this case we are getting access tokens for the Build API, let's be direct about it in the commit message, e.g: `Get access token for Build API using JWT`.

pkg/client/client.go
```diff
@@ -31,7 +31,8 @@ import (
 const (
 	// Value to pass as credentials to the Host Orchestrator service endpoints. Any non-empty value is enough.
 	InjectedCredentials             = "inject"
-	headerNameCOInjectBuildAPICreds = "X-Cutf-Cloud-Orchestrator-Inject-BuildAPI-Creds"
+	HeaderNameCOInjectBuildAPICreds = "X-Cutf-Cloud-Orchestrator-Inject-BuildAPI-Creds"
+	HeaderNameHOBuildAPICreds       = "X-Cutf-Host-Orchestrator-BuildAPI-Creds"
```
ser-io, line 35: Use existing `const` https://github.com/google/cloud-android-orchestration/blob/main/pkg/client/host_orchestrator_client.go#L70, need to make it public though.

pkg/client/client.go
```diff
@@ -189,7 +191,7 @@ func (s *serviceImpl) HostService(host string) HostOrchestratorService {
 	hs := &HostOrchestratorServiceImpl{
 		HTTPHelper: s.httpHelper,
 		// Make the cloud orchestrator inject the credentials instead
-		BuildAPICredentialsHeader: headerNameCOInjectBuildAPICreds,
+		BuildAPICredentialsHeader: s.CredHeaderName,
```
ser-io, line 194: default to `X-Cutf-Host-Orchestrator-BuildAPI-Creds` here, and modify it down below with `headerNameCOInjectBuildAPICreds` if s.InjectBuildAPICreds is true.

ser-io (changes requested):

ser-io (approved):

PR #349, opened 2024-08-21
Env config user artifacts

ser-io:


pkg/cli/cvd.go
```diff
@@ -159,6 +162,50 @@ func (c *cvdCreator) Create() ([]*hoapi.CVD, error) {
...
+	}
+	var b bytes.Buffer
+	err = tmpl.Execute(&b, uaEnvConfigTmplData{ArtifactsDir: d.ArtifactsDir})
+	if err != nil {
+		panic(err)
+	}
+	result := make(map[string]interface{})
+	if err := json.Unmarshal(b.Bytes(), &result); err != nil {
+		panic(err)
+	}
```
0405ysj, line 205: Uniting the style for error handling? Former reuses `err` variable, but latter declares `err` variable inside conditional statement.

pkg/cli/cvd.go
```diff
@@ -159,6 +162,50 @@ func (c *cvdCreator) Create() ([]*hoapi.CVD, error) {
...
+`
+
+type uaEnvConfigTmplData struct {
+	ArtifactsDir string
+}
+
+func (d *uaEnvConfigTmplData) Value() map[string]interface{} {
+	tmpl, err := template.New("").Parse(uaEnvConfigTmpl)
+	if err != nil {
+		panic(err)
```
0405ysj, line 195: Why does this function use `panic` to handle error, though all callers of this function already does error propagation?

0405ysj (approved):
Left some comments, but comments doesn't harm entire code design of this PR. Pressing approval in advance.

pkg/cli/cvd.go
ser-io: my thinking was that an error parsing the template should not be propagated, however the other two errors in this method should be propagated. I moved parsing the template into an `init` block.

PR #351, opened 2024-08-22
 Discard X-Cutf-Cloud-Orchestrator-Inject-BuildAPI-Creds header 

ser-io:


pkg/app/app.go
```diff
@@ -148,8 +148,8 @@ func (a *App) InfraConfig() apiv1.InfraConfig {
 }
 
 const (
-	headerNameCOInjectBuildAPICreds = "X-Cutf-Cloud-Orchestrator-Inject-BuildAPI-Creds"
-	headerNameHOBuildAPICreds       = "X-Cutf-Host-Orchestrator-BuildAPI-Creds"
+	headerNameHOBuildAPICreds = "X-Cutf-Host-Orchestrator-BuildAPI-Creds"
```
jemoreira, line 151: This is an API breaking change, why rename the header?

jemoreira (approved):

pkg/app/app.go
ser-io: discussed offline, let's keep the two-header solution. will be closing the PR

PR #354, opened 2024-08-29
Change cvdr websocket connect agent to follow cvdr command scheme

k311093:
Changed to use --zone and --host flag to find adb WebSocket backend.

When --host=none, use service_url as before (Directly connect to service_url) to support host orchestrator only case.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -1319,6 +1319,16 @@ func runConnectionWebrtcAgentCommand(flags *ConnectFlags, c *command, args []str
 	return nil
 }
 
+func wsURL(wsRoot string, zone string, host string, device string) string {
+	var b strings.Builder
+	b.WriteString(wsRoot)
+	if host != "none" {
+		b.WriteString(fmt.Sprintf("/v1/zones/%s/hosts/%s", zone, host))
+	}
+	b.WriteString(fmt.Sprintf("/devices/%s/adb", device))
+	return b.String()
+}
+
 func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []string, opts *subCommandOpts) error {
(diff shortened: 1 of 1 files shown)


0405ysj (approved):

pkg/cli/cli.go
```diff
@@ -1358,7 +1358,13 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
 		},
 	}
 	device := args[0]
-	wsUrl := fmt.Sprintf("%s/devices/%s/adb", wsRoot, device)
+	var wsUrlBuilder strings.Builder
```
ser-io, line 1361: let's isolate this logic into a new function `wsURL()`.

pkg/cli/cli.go
```diff
@@ -1358,7 +1358,13 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
...
+	wsUrlBuilder.WriteString(wsRoot)
```
ser-io, line 1362: [docker-image-v0.1.0-alpha](https://google.github.io/styleguide/go/decisions.html#initialisms) In go, acronyms should have the same case, it should be `wsURLBuilder` in this case. However in the new function `wsURL()` you can have shorter names and avoid the `URL` part as it's going to be redundant.

ser-io (approved):

pkg/cli/cli.go
k311093: Extracted the logic into the function wsURL(...).

pkg/cli/cli.go
k311093: Changed to name to `url` to follow acronyms case rule and avoid confusion with the function name wsURL().

PR #358, opened 2024-09-06
Align WebSocket connection request with normal HTTP request at cvdr

k311093:
WebSocket connection needs to follow Proxy configuration and Authorization configuration to connect adb websocket backend through Cloud Orchestrator. Change to use same config and header as normal HTTP request does.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -1319,26 +1319,35 @@ func runConnectionWebrtcAgentCommand(flags *ConnectFlags, c *command, args []str
 	return nil
 }
 
-func wsURL(wsRoot string, zone string, host string, device string) string {
-	var b strings.Builder
-	b.WriteString(wsRoot)
-	if host != "none" {
-		b.WriteString(fmt.Sprintf("/v1/zones/%s/hosts/%s", zone, host))
+func connectADBWebSocketDirect(serviceURL string, device string) (*websocket.Conn, error) {
+	// Connect to ADB proxy WebSocket directly by using serviceURL as Operator URL.
+	//   wss://127.0.0.1:1443/devices/cvd-1/adb
+	url, err := url.Parse(serviceURL)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to parse URL %s: %w", serviceURL, err)
 	}
-	b.WriteString(fmt.Sprintf("/devices/%s/adb", device))
-	return b.String()
+	switch p := &url.Scheme; *p {
+	case "https":
+		*p = "wss"
+	case "http":
+		*p = "ws"
+	default:
+		return nil, fmt.Errorf("Unknown scheme %s", *p)
+	}
+	url = url.JoinPath("devices", device, "adb")
+	dialer := websocket.Dialer{
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: true,
+		},
+	}
+	wsConn, _, err := dialer.Dial(url.String(), nil)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to connect WebSocket %s: %w", url.String(), err)
+	}
+	return wsConn, nil
 }
 
 func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []string, opts *subCommandOpts) error {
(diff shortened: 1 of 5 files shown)


ikicha (approved):
Great! Thanks.

pkg/client/client.go
```diff
@@ -86,6 +86,8 @@ type Service interface {
 	HostService(host string) HostOrchestratorService
 
 	RootURI() string
+
+	AuthHeader() (http.Header, error)
```
jemoreira, line 90: The code that establishes the websocket connection should be part of the `HostOrchestratorService`, so that it can be used from places other than the cli.

With that change there would be no need to expose the `AuthHeader` in the CO client, it can simply pass it to the HO client behind the scenes.

I understand this is a big change, so it's ok to do it in another PR.

0405ysj (approved):
Thanks!

pkg/client/client.go
k311093: Implemented WebSocket connection at the HO client and changed to use it when possible.

Thanks!

ser-io (changes requested):
I think we can move the `PrepareAuth` refactoring into the last commit and have only one commit in this PR.

Having 3 commits were most of the changes in commit 1 and commit 2 are undo in commit 3 is misleading when looking at the history as well as getting the full picture of this PR.

k311093:
> I think we can move the `PrepareAuth` refactoring into the last commit and have only one commit in this PR.
> 
> Having 3 commits were most of the changes in commit 1 and commit 2 are undo in commit 3 is misleading when looking at the history as well as getting the full picture of this PR.

Merged into one commit.

pkg/client/host_orchestrator_client.go
```diff
@@ -114,6 +122,43 @@ func (c *HostOrchestratorServiceImpl) ConnectWebRTC(device string, observer wcli
...
+	}
+	if c.ProxyURL != "" {
+		proxyURL, err := url.Parse(c.ProxyURL)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to parse proxy %s: %w", c.ProxyURL, err)
+		}
+		dialer.Proxy = http.ProxyURL(proxyURL)
+	}
+
+	wsRoot := c.HTTPHelper.RootEndpoint
```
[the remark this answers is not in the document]
k311093: Changed to use URL object. Also used the object to join URL path.

pkg/client/httputils.go
```diff
@@ -187,17 +187,25 @@ func (rb *HTTPRequestBuilder) JSONResDoWithRetries(ret any, retryOpts RetryOptio
 	return apiError
 }
 
-func (rb *HTTPRequestBuilder) doWithRetries(retryOpts RetryOptions) (*http.Response, error) {
+func (rb *HTTPRequestBuilder) PrepareAuth() error {
```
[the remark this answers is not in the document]
k311093: Changed function name into setAuthz.

ser-io:
> > I think we can move the `PrepareAuth` refactoring into the last commit and have only one commit in this PR.
> > Having 3 commits were most of the changes in commit 1 and commit 2 are undo in commit 3 is misleading when looking at the history as well as getting the full picture of this PR.
> 
> Merged into one commit.

Good work. Thanks!

ser-io (approved):

pkg/cli/cli.go
```diff
@@ -1319,26 +1319,36 @@ func runConnectionWebrtcAgentCommand(flags *ConnectFlags, c *command, args []str
...
+			InsecureSkipVerify: true,
+		},
 	}
-	b.WriteString(fmt.Sprintf("/devices/%s/adb", device))
-	return b.String()
+	wsConn, _, err := dialer.Dial(url.String(), nil)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to connect WebSocket %s: %w", url.String(), err)
+	}
+
```
ser-io, line 1347: nit: remove empty line.

ser-io (approved):

pkg/cli/cli.go
k311093: Removed empty line.

Also removed other empty lines.

PR #371, opened 2024-10-28
[Code Health] Rename and cleanup.

ser-io:


pkg/cli/cli.go
```diff
@@ -1578,6 +1590,9 @@ func newClient(config Config, flags *ServiceFlags, c *cobra.Command) (client.Cli
 	if flags.ServiceURL == unitTestServiceURL {
 		return &fakeClient{}, nil
 	}
+	if flags.ServiceURL == client.LocalDockerServiceURL {
+		return client.NewDockerClient(&client.ClientOptions{})
+	}
```
jemoreira, line 1595: Why is a different client implementation needed for local docker?

The different behavior should be implemented in the cloud orchestrator, with the client simply accessing its api in the same way other it does for other services.

jemoreira:
More information in the PR name and description would be greatly appreciated

pkg/cli/cli.go
ser-io: When used locally, rather than running the Cloud Orchestrator server locally and interact with it through the http api, we can interact with the Docker Engine directly from `cvdr` this makes it easier to manage cuttlefish instances when used locally. It's also easier for folks to reproduce the same environment they don't need to bother with running/maintaining a local Cloud Orchestrator service.

ser-io:
> More information in the PR name and description would be greatly appreciated

Changed the PR description and added more context to the commit implementing the Local Docker Client https://github.com/google/cloud-android-orchestration/pull/371/commits/56052638783d5ec7e082d3acc3020572e2084de1

pkg/cli/cli.go
```diff
@@ -826,11 +824,15 @@ func runCreateCVDCommand(c *cobra.Command, args []string, flags *CreateCVDFlags,
...
+	hostSrvURL, err := srvClient.HostServiceURL(flags.CreateCVDOpts.Host)
+	if err != nil {
+		return fmt.Errorf("failed getting host service url: %w", err)
+	}
 	hosts := []*RemoteHost{
 		{
-			ServiceRootEndpoint: srvClient.RootURI(),
-			Name:                flags.CreateCVDOpts.Host,
-			CVDs:                cvds,
+			ServiceURL: hostSrvURL,
```
0405ysj, line 833: I prefer `HostServiceURL` rather than `ServiceURL` in perspective of code readers. AFAIU, we're using the term `ServiceURL` as the address of Cloud Orchestration widely.

pkg/cli/testing.go
```diff
@@ -34,11 +34,11 @@ func (fakeClient) RootURI() string {
 	return unitTestServiceURL + "/v1"
 }
 
-func (fakeClient) HostService(host string) hoclient.HostOrchestratorService {
+func (fakeClient) HostService(host string) (hoclient.HostOrchestratorService, error) {
 	if host == "" {
 		panic("empty host")
```
0405ysj, line 39: Probably you want to return an error than occurring panic?

pkg/client/localdocker.go
```diff
@@ -0,0 +1,212 @@
...
+		PortBindings: nat.PortMap{hoSrvPort: []nat.PortBinding{{HostPort: ""}}},
+	}
+	res, err := s.dcli.ContainerCreate(ctx, config, hostConfig, nil, nil, "")
+	if err != nil {
+		return nil, fmt.Errorf("container create failed: %w", err)
+	}
+	err = s.dcli.ContainerStart(ctx, res.ID, types.ContainerStartOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("container start failed: %w", err)
+	}
```
0405ysj, line 76: I think lots of code in this file is duplicated to `pkg/app/instances/docker.go`, and I think it's better to make library which both `pkg/app/instances/docker.go` and `pkg/client/localdocker.go` can use. I'm little afraid of the execution flow like container config between remote docker and local docker being distinguished even for unexpected ones in the future.

0405ysj: Or, is there any reason for not making such library?

0405ysj (changes requested):

pkg/cli/cli.go
```diff
@@ -1578,6 +1590,9 @@ func newClient(config Config, flags *ServiceFlags, c *cobra.Command) (client.Cli
 	if flags.ServiceURL == unitTestServiceURL {
 		return &fakeClient{}, nil
 	}
+	if flags.ServiceURL == client.LocalDockerServiceURL {
```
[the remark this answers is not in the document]
ser-io: correct, I didn't want to introduce a new flag in order to start using this new functionality. I agree about documentation, however the feature is experimental for now, meant to be used as a helper to develop `cvdr` and `host orchestrator`, not meant for every user yet. We don't have a guide for development yet, which is where we can add documentation for it.

pkg/cli/cli.go
ser-io: Good point, I thought about it as well, decided to stick with `ServiceURL` as the `Host` prefix would be redundant as it's a field of the `RemoteHost` struct. 

````
host := RemoteHost{...}` 
host.ServiceURL // I think it's clear in this case this is the url of the host orchestrator api
```

pkg/client/localdocker.go
ser-io: there's some duplicated code, however, let's not constrain ourselves by having a rushed refactoring at this point and have the ability of using the docker library independently as there are some key difference in its use, for example: host port mapping and the using/not using the Operations object.

pkg/cli/testing.go
ser-io: Done.

ser-io:
After discussing with @jemoreira, dropped the client implementation for local docker in favor of making it easier to run the Cloud Orchestrator with Docker Instance Management for development. I'll keep the previous commits in the PR and re-brand the PR as a code health effort

pkg/client/localdocker.go
ser-io: no longer relevant.

pkg/cli/cli.go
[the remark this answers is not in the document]
ser-io: no longer relevant

pkg/cli/cli.go
ser-io: no longer relevant.

pkg/cli/cli.go
0405ysj: Understood. It sounds fair to me.

0405ysj (approved):
Approval from myself, but you probably want another approval from @jemoreira.

jemoreira (approved):

PR #374, opened 2024-11-08
Create remote devices using environment configs pointing to local art…

ser-io:
…ifacts.

- `cvdr` can be used with environment configs like the following where common.host_package and instances[0].disk.default_build points to local artifacts.

```
{
  "common": {
    "host_package": "/tmp/artifacts/cvd-host_package.tar.gz"
  },
  "instances": [
    {
      "vm": {
        "memory_mb": 8192,
        "setupwizard_mode": "OPTIONAL",
        "cpus": 8
      },
      "disk": {
        "default_build": "/tmp/artifacts/images.zip"
      },
      "streaming": {
        "device_id": "cvd-1"
      }
    }
  ]
}
```

### DIFF
--- a/pkg/cli/cvd.go
+++ b/pkg/cli/cvd.go
@@ -99,8 +99,7 @@ type CreateCVDOpts struct {
 	LocalImage      bool
 	// Creates multiple instances. Only relevant if given a single build source.
 	NumInstances int
-	// Structure: https://android.googlesource.com/device/google/cuttlefish/+/8bbd3b9cd815f756f332791d45c4f492b663e493/host/commands/cvd/parser/README.md
-	// Example: https://cs.android.com/android/platform/superproject/main/+/main:device/google/cuttlefish/host/cvd_test_configs/main_phone-main_watch.json;drc=b2e8f4f014abb7f9cb56c0ae199334aacb04542d
+	// TODO(b/378123925): Work with https://github.com/google/android-cuttlefish/blob/main/base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_config.proto
 	EnvConfig map[string]interface{}
 	// If true, perform the ADB connection automatically.
 	AutoConnect               bool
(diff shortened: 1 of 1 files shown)


pkg/cli/cvd.go
```diff
@@ -653,3 +738,22 @@ func uploadFiles(srv hoclient.HostOrchestratorService, uploadDir string, names [
 	}
 	return nil
 }
+
+// Deep copies src to dst using json marshaling.
+func deepCopy(src, dst any) {
+	b, err := json.Marshal(src)
+	if err != nil {
+		panic(err)
```
jemoreira, line 746: Don't panic inside these utility functions, return an error instead and let the cli-specific code decide how to handle that (possibly panicking).

pkg/cli/cvd.go
```diff
@@ -278,18 +278,103 @@ func (c *cvdCreator) createCVDFromAndroidCI() ([]*hoapi.CVD, error) {
...
 
+func (c *cvdCreator) uploadFilesFromEnvConfig(hostSrv hoclient.HostOrchestratorService, config map[string]interface{}) error {
+	if err := c.uploadCVDHostPackageFromEnvConfig(hostSrv, config); err != nil {
+		return err
+	}
+	return c.uploadImagesFromEnvConfig(hostSrv, config)
+}
+
+func (c *cvdCreator) uploadCVDHostPackageFromEnvConfig(hostSrv hoclient.HostOrchestratorService, config map[string]interface{}) error {
+	common, ok := config["common"]
```
jemoreira, line 307: There is a type-safe definition of what a config file contains (using protobuffer) in the android-cuttlefish project. Consider using that instead of accessing the JSON file through map keys.

pkg/cli/cvd.go
```diff
@@ -278,18 +278,103 @@ func (c *cvdCreator) createCVDFromAndroidCI() ([]*hoapi.CVD, error) {
...
-	res, err := c.client.HostService(c.opts.Host).CreateCVD(createReq, c.credentialsFactory())
+	res, err := hostSrv.CreateCVD(createReq, c.credentialsFactory())
 	c.statePrinter.PrintDone(stateMsgFetchAndStart, err)
 	if err != nil {
 		return nil, err
 	}
 	return res.CVDs, nil
 }
 
+func (c *cvdCreator) uploadFilesFromEnvConfig(hostSrv hoclient.HostOrchestratorService, config map[string]interface{}) error {
```
jemoreira, line 299: This function doesn't just upload the files, but also updates the config. Please rename to make this fact more evident.

pkg/cli/cvd.go
```diff
@@ -278,18 +278,103 @@ func (c *cvdCreator) createCVDFromAndroidCI() ([]*hoapi.CVD, error) {
 }
 
 func (c *cvdCreator) createWithCanonicalConfig() ([]*hoapi.CVD, error) {
+	hostSrv := c.client.HostService(c.opts.Host)
+	envConfig := make(map[string]any)
+	deepCopy(c.opts.EnvConfig, &envConfig)
```
jemoreira, line 283: Why make a copy? Is `c.opts.EnvConfig` used for anything else after this?

pkg/cli/cvd.go
ser-io: Done.

pkg/cli/cvd.go
ser-io: Good point, created b/378123925 and added relevant TODOs in the code to work with the protobuf definition.  However, I'd like to do this in a follow up PR as this would require some work in the `android-cuttlefish` repo as well to make the proto available as a Golang library.

pkg/cli/cvd.go
ser-io: I don't have a strong opinion here, I'm just biased on keeping a reference to the original config through out the whole execution.

pkg/cli/cvd.go
ser-io: Good point. Done.

pkg/cli/cvd.go
jemoreira: Sounds good

jemoreira (approved):

PR #376, opened 2024-11-20
feat: add Cloud Run deployment

kunzese:
This PR adds the Terraform setup and the necessary changes to Cloud Orchestrator to be deployed on Cloud Run protected by IAP. Additionally it includes the possibility to use GCE instances with private IPs only.

Co-authored-by: David Gleich <gleichda@google.com>

### DIFF
--- a/pkg/app/instances/gce.go
+++ b/pkg/app/instances/gce.go
@@ -38,7 +38,7 @@ type GCPIMConfig struct {
 	HostOrchestratorPort int
 	Network              string
 	Subnetwork           string
-	UsePrivateIP         bool
+	UseExternalIP        bool
 	// If true, instances created should be compatible with `acloud CLI`.
 	AcloudCompatible bool
 }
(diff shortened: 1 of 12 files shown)


kunzese:
@gleichda FYI

scripts/gcp/app/conf.toml.tmpl
```diff
@@ -20,7 +20,7 @@
 WebStaticFilesPath = "web"
 
 [AccountManager]
-Type = "GCP"
+Type = "IAP"
```
ser-io, line 23: Let's keep App Engine deployment config with `GCP` for know as there's no documentation to enable IAP for the App Engine deployment publicly available yet.

terraform/.terraform.lock.hcl
ser-io, line 1: is this file needed in the repo?

terraform/README.md
ser-io, line 1: let's move the new `terraform` directory under `scripts/gcp/cloudrun/`

ser-io (changes requested):

terraform/.terraform.lock.hcl
kunzese: Yes i is advised to commit this file too. It is like the `go.sum` file a record of all dependencies and their versions.

pkg/app/instances/gce.go
```diff
@@ -38,6 +38,7 @@ type GCPIMConfig struct {
 	HostOrchestratorPort int
 	Network              string
 	Subnetwork           string
+	UsePrivateIP         bool
```
ser-io, line 41: let's default to private ip only, and add the config parameter to enabling external ip.

ser-io (approved):

ser-io (approved):

ser-io (changes requested):
Rebase the `cloudrun-clean` branch on top of main instead rather than having a merge commit.

PR #377, opened 2024-11-21
Allow to specify boot disk size for GCE instances and auto delete boot disk

gleichda:
- Add option for GCE To specify boot disk size; It will fall back to the image size if unset (Keeping the same default)
- Move the `BootDiskSizeGB` to be GCP Only as it is used nowhere else and is GCP specific
- Configure Boot disks to auto delete to properly clean up

New config would look like:
```
SystemDefaultService = "dockerIM"
[Services.dockerIM]
ServiceURL = "https://myservice.example.com"
Zone = "europe-west3-a"

Host={
	GCP={
		MachineType = "n1-standard-4",
		MinCPUPlatform = "Intel Haswell",
		BootDiskSizeGB = 100
	}
}
Authn={
	OIDCToken={
		TokenFile = "/tmp/cvdr_oidc.token"
	}
}
```

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -86,12 +86,14 @@ const (
 
 	gcpMachineTypeFlag    = "gcp_machine_type"
 	gcpMinCPUPlatformFlag = "gcp_min_cpu_platform"
+	gcpBootDiskSizeGB     = "gcp_boot_disk_size_gb"
 )
 
 const (
 	acceleratorFlagDesc       = "Configuration to attach accelerator cards, i.e: --accelerator type=nvidia-tesla-p100,count=1"
 	gcpMachineTypeFlagDesc    = "Indicates the machine type"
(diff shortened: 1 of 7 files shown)


ser-io (changes requested):
let's split this PR into two commits: one for auto delete boot disk and the other for boot disk size.

gleichda:
Tested the performance impact of setting the disk size via Cloud Orchestrator and the additional startup latency is negligible
image

pkg/cli/cli.go
```diff
@@ -86,12 +86,14 @@ const (
 
 	gcpMachineTypeFlag    = "gcp_machine_type"
 	gcpMinCPUPlatformFlag = "gcp_min_cpu_platform"
+	gcpBootDiskSizeGB 	  = "gcp_boot_disk_size_gb"
```
ser-io, line 89: nit: run go lint `gofmt`

ser-io (approved):

PR #381, opened 2024-12-09
Add client.WithAPIVersionNegotiation() option for Docker client

k311093:
This makes Cloud Orchestrator possible to connect to docker servie with older version than docker client library.

### DIFF
--- a/cmd/cloud_orchestrator/main.go
+++ b/cmd/cloud_orchestrator/main.go
@@ -63,7 +63,7 @@ func LoadInstanceManager(config *config.Config) instances.Manager {
 	case instances.UnixIMType:
 		im = instances.NewLocalInstanceManager(config.InstanceManager)
 	case instances.DockerIMType:
-		cli, err := client.NewClientWithOpts(client.FromEnv)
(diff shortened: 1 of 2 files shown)


ikicha (approved):

0405ysj:
Can we get rid of code around calculating docker API version in `scripts/docker/run.sh`?

0405ysj (approved):

k311093:
> Can we get rid of code around calculating docker API version in `scripts/docker/run.sh`?

I think so, it might do similar things at `scripts/docker/run.sh`. I tried just running cloud_orchestrator with configuration and succeeded to run with older docker version without additional environments.

0405ysj:
> I think so, it might do similar things at scripts/docker/run.sh

Then, why don't we cleanup that code with applying this? BTW, I'm okay to cleaning it up in following PR.

k311093:
> > I think so, it might do similar things at scripts/docker/run.sh
> 
> Then, why don't we cleanup that code with applying this? BTW, I'm okay to cleaning it up in following PR.

Added commit for removing docker version related logic for script/docker/run.sh

jemoreira (approved):

PR #383, opened 2024-12-11
Write ping message for preventing idle timeout for ADB websocket

k311093:
WebSocket connection is closed when it is idle for a while. Write ping message for every 10 seconds to prevent idle timeout for ADB websocket connection.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -30,6 +30,7 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+	"time"
 
 	client "github.com/google/cloud-android-orchestration/pkg/client"
 
@@ -1534,11 +1535,7 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
 	}
 
 	// Redirect WebSocket to ADB TCP port
-	wsWrapper := &wsIoWrapper{
-		wsConn: wsConn,
-		pos:    0,
-		buf:    nil,
-	}
+	wsWrapper := newWsIoWrapper(c, wsConn)
 	go func() {
 		io.Copy(wsWrapper, tcpConn)
 		wsWrapper.Close()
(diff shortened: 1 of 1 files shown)


pkg/cli/cli.go
```diff
@@ -1423,6 +1424,20 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
...
 
+	go func() {
+		ticker := time.NewTicker(10 * time.Second)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ticker.C:
+				if err := wsConn.WriteControl(websocket.PingMessage, []byte{}, time.Now().Add(5*time.Second)); err != nil {
+					c.PrintErrf("Failed to write Ping: %v", err)
+					return
```
jemoreira, line 1435: IIUC this will only exit when there is an error. There should be a mechanism to do an orderly shutdown of this routine. For example, there could be another channel in the `select` call that simply causes it to return on any message received. The channel would be written to right before closing the `wsWrapper` in line 1450.

k311093: Added channel for the scenario that socket is shutdown normally.

Moved go routine code into newWsIoWrapper function that is currently introduced for create wsIoWrapper.

ikicha (approved):

jemoreira (approved):

pkg/cli/cli.go
```diff
@@ -1439,13 +1436,40 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
...
+	pos      int
+	buf      []byte
+	stopPing chan interface{}
 }
 
 var _ io.ReadWriteCloser = (*wsIoWrapper)(nil)
 
+func newWsIoWrapper(c *command, wsConn *websocket.Conn) *wsIoWrapper {
+	stopPing := make(chan interface{})
+	go func(stopPing chan interface{}) {
```
ser-io, line 1449: let's move this logic out of the `newWsIoWrapper` constructor. 

Suggest having this logic around a new struct: `wsPingSender`

```
type wsPingSender struct {
  wsConn *websocket.Conn
  delay time.Duration
  stopCh chan interface{}
}

func (w *wsPingSender) Start() error {
 // Add logic here
}
```

ser-io (changes requested):

pkg/cli/cli.go
```diff
@@ -1439,13 +1436,40 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
...
+		ticker := time.NewTicker(10 * time.Second)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ticker.C:
+				if err := wsConn.WriteControl(websocket.PingMessage, []byte{}, time.Now().Add(5*time.Second)); err != nil {
+					c.PrintErrf("Failed to write Ping: %v", err)
```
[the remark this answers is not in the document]
k311093: Changed message to `ping message`

pkg/cli/cli.go
k311093: Declared new type `wsPingSender` and changed `wsIoWrapper` to use it.

ikicha (approved):

pkg/cli/cli.go
ser-io: Nice! thanks.

ser-io (approved):

PR #384, opened 2024-12-11
Cloud Orchestrator can download docker image from Artifact Registry

0405ysj:


pkg/app/instances/docker.go
```diff
@@ -28,22 +31,24 @@ import (
...
 	"github.com/docker/docker/api/types/filters"
+	"github.com/docker/docker/api/types/image"
 	"github.com/docker/docker/client"
 )
 
 const DockerIMType IMType = "docker"
 
 type DockerIMConfig struct {
-	DockerImageName      string
+	DockerImageNames     []string
```
ser-io, line 41: add documentation for this new config. why having many images, which is supposed to be the one host needs to be created with?

pkg/app/instances/docker.go
```diff
@@ -53,11 +58,63 @@ const (
 	DeleteHostOPType OPType = "deletehost"
 )
 
-func NewDockerInstanceManager(cfg Config, cli *client.Client) *DockerInstanceManager {
-	return &DockerInstanceManager{
-		Config: cfg,
-		Client: cli,
+func NewDockerInstanceManager(cfg Config, cli *client.Client) (*DockerInstanceManager, error) {
+	ctx := context.TODO()
```
ser-io, line 62: The new logic to manage images should be when creating the hosts (docker containers) rather than when creating an instance of the `DockerInstanceManager`.

ser-io (changes requested):

pkg/app/instances/docker.go
0405ysj: Done.

pkg/app/instances/docker.go
0405ysj: Could you elaborate about why you prefer this way? IMO, docker instance manager without any docker image cannot work its role properly.

ser-io: It's about consistency with other instance manager implementations like GCP, if we use an invalid GCP image resource name, we are still able to launch the cloud orchestrator binary,  we fail  with error `500 Internal Server Error` Image Not Found, but we don't prevent the Cloud Orchestrator binary to start running. 

Making this logic part of the constructor makes the binary to crash at the beginning, the service won't ever start. If we restart the orchestrator in a machine that already have hosts running in a new version deployment, by having any issue in this new logic in the constructor the binary will crash and never start, hence we cannot even list and connect with the existing hosts, we should only be affecting the create logic, but not the whole service.

0405ysj: Understood, and it sounds fair. I wasn't aware of such history on GCE impl.

0405ysj: Done.

pkg/app/instances/docker.go
ser-io: I don't understand the documentation.

>If 2 or more images exist, former one would be chosen. 

If we chose the first one, why having a list of images?

> If no images exist, downloadable image with shortest latency would be chosen.

What do you mean with  "If no images exist"? Do you mean the list is empty or there's no relevant image in the local Docker Engine? Elaborate on what's a downloadable image.

Having a single bucket for image names, seems to make these logic complicated, feel free to add a new config parameter in order to make it easier to follow.

pkg/app/instances/docker.go
```diff
@@ -297,3 +312,51 @@ func (m *DockerInstanceManager) getContainerLabel(host string, key string) (stri
...
+		for _, existingImage := range listRes {
+			for _, existingTag := range existingImage.RepoTags {
+				if candidateImage == existingTag {
+					return candidateImage, nil
+				}
+			}
+		}
+	}
+
+	var chosenImage string
```
ser-io, line 331: This logic to determine the fastest source is unreliable as you're making just one request which is not enough to describe latency, we would need a bigger sample. Let's have a configuration value for this and let the user change it depending on its own deployment.

ser-io (changes requested):

pkg/app/instances/docker.go
0405ysj: I had discussion about this with @ikicha, and decided to remove this logic and not bringing new configuration field `DockerImageNames`. `conf.toml` would be kept as current, and deployment of `conf.toml` to Artifact Registry would be happened by modifying the value of `conf.toml`.

pkg/app/instances/docker.go
```diff
@@ -28,17 +31,23 @@ import (
...
-	DockerImageName      string
+	// Candidates of docker image names.
+	// If 2 or more images exist, former one would be chosen.
+	// If no images exist, downloadable image with shortest latency would be chosen.
+	DockerImageNames     []string
 	HostOrchestratorPort int
 }
 
 const dockerLabelCreatedBy = "created_by"
+const dockerLabelKeyManagedBy = "managed_by"
```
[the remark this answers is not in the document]
0405ysj: Done.

pkg/app/instances/docker.go
0405ysj: Removed that logic.

pkg/app/instances/docker.go
0405ysj: Canceled to apply `DockerImageNames` configuration.

pkg/app/instances/docker.go
```diff
@@ -297,3 +310,26 @@ func (m *DockerInstanceManager) getContainerLabel(host string, key string) (stri
...
+				return nil
+			}
+		}
+	}
+
+	out, err := m.Client.ImagePull(ctx, m.Config.Docker.DockerImageName, image.PullOptions{})
+	if err != nil {
+		return fmt.Errorf("Failed to pull docker image '%s': %w", m.Config.Docker.DockerImageName, err)
+	}
+	defer out.Close()
```
ikicha, line 331: why do we need `out` and copying? can we just ignore the first return value of `ImagePull`?

0405ysj: I thought so, but `ImagePull` API doesn't work synchronously. Without handling `out`, we cannot guarantee whether the image is pulled or not.

https://docs.docker.com/reference/api/engine/sdk/examples/

ikicha (approved):

ser-io (changes requested):

pkg/app/instances/docker.go
```diff
@@ -297,3 +310,26 @@ func (m *DockerInstanceManager) getContainerLabel(host string, key string) (stri
...
+	io.Copy(io.Discard, out)
```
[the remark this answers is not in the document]
0405ysj: done.

pkg/app/instances/docker.go
0405ysj: done.

ikicha (approved):

pkg/app/instances/docker.go
```diff
@@ -297,3 +310,31 @@ func (m *DockerInstanceManager) getContainerLabel(host string, key string) (stri
...
+	reader, err := m.Client.ImagePull(ctx, m.Config.Docker.DockerImageName, image.PullOptions{})
+	if err != nil {
+		return fmt.Errorf("Failed to request pulling docker image '%s': %w", m.Config.Docker.DockerImageName, err)
+	}
+	defer reader.Close()
+	// Caller of ImagePull should handle its output to complete actual ImagePull operation.
+	// Details in https://pkg.go.dev/github.com/docker/docker/client#Client.ImagePull.
+	_, err = io.Copy(io.Discard, reader)
+	if err != nil {
+		return fmt.Errorf("Failed to pull docker image '%s': %w", m.Config.Docker.DockerImageName, err)
```
ser-io, line 336: nit: IIRC if using format verb `%q` rather than `%s` you don't have to add the quotes yourself.

ser-io (approved):

pkg/app/instances/docker.go
0405ysj: Done, thanks for the review!

PR #386, opened 2024-12-16
Fix many golint errors

0405ysj:
Removed all errors shown in: \
`$ golint -min_confidence 0 ./... | grep -v comment | grep -v api/v1`

Reasoning:
- Fixing lint errors for writing comment takes too much effort.
- Fixing lint errors in `api/v1/*` may break API surface.

### DIFF
--- a/pkg/app/accounts/accounts.go
+++ b/pkg/app/accounts/accounts.go
@@ -35,5 +35,5 @@ type AMType string
 
 type Config struct {
 	Type   AMType
-	OAuth2 appOAuth2.OAuth2Config
+	OAuth2 appOAuth2.Config
 }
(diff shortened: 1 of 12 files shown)


ser-io (changes requested):
Thanks for taking this initiative.

 Use https://staticcheck.dev/ as golint is [deprecated and frozen](https://github.com/golang/lint)

Let's also avoid having all the checks fixed in one commit, it's problematic in case we need to revert the commit, Let's group check together and have smaller commits, e.g we can have one commit only fixing check [ST1005] (https://staticcheck.dev/docs/checks/#ST1005) `staticcheck -checks=ST1005 ./...`

See PR https://github.com/google/android-cuttlefish/pull/796 as a reference.

0405ysj:
Acknowledged. Closing all PRs about golint, will prepare with staticcheck instead.

PR #387, opened 2024-12-16
Golint presubmit

0405ysj:
This kind of PR should be landed into `android-cuttlefish` repo as well. Also please ignore presubmit check failure now, since it's intended, will be fixed after merging https://github.com/google/cloud-android-orchestration/pull/386.


0405ysj:
Acknowledged. Closing all PRs about golint, will prepare with staticcheck instead.

PR #391, opened 2025-01-06
Refactor to distinguish logic calling CreateCVD HO API

0405ysj:


ikicha:
can you provide the context for this?

0405ysj:
This change is for supporting cloning host without upload artifacts for many times, but currently not sure to land this PR due to the comments in the pending design. Let me make this PR as draft again.

ser-io (changes requested):
Please read [go/tott/703](https://goto.google.com/tott/703). This refactoring is not helpful as we don't have a clear abstraction yet for creating devices in different ways. Let's not constrain the different create flows for now.

Adding `hostService hoclient.HostOrchestratorService` field to `cvdCreator` is positive.

PR #394, opened 2025-01-09
Enable using staticcheck in this repository

0405ysj:


pkg/app/accounts/unix.go
```diff
@@ -17,18 +17,13 @@ package accounts
...
-
-	"golang.org/x/oauth2"
 )
 
 const UnixAMType AMType = "unix"
 
 // Implements the Manager interface taking the username from the
 // environment and authorizing all requests
-type UnixAccountManager struct {
-	tokenSource oauth2.TokenSource
```
ikicha, line 29: @ser-io @jemoreira can you confirm that we don't need them anymore?

ser-io: yes, these fields should have been deleted in https://github.com/google/cloud-android-orchestration/commit/d171d35811a20f8bd3fca5fd3f37cbd59bcfcd2f

ser-io (approved):
Thanks.

Issue #382, opened 2024-12-10
solve webrtc do not support turn server, it can not use from diffrent computer


l1351868270:
I start an cuttlefish flow the doc, below
```
./scripts/docker/run.sh

CVDR_USER_CONFIG_PATH=scripts/docker/cvdr.toml ./cvdr create --local_cvd_host_pkg_src=/data/lsl/android14-qpr1-release/out/dist/cvd-host_package.tar.gz --local_images_zip_src=/data/lsl/android14-qpr1-release/out/dist/aosp_cf_x86_64_phone-img-eng.sun.zip --auto_connect=false
```

I want to open the http://xxx.xxx.xxx.xxx:8080/v1/zones/local/hosts/0b765db27cfcaebdb3d939a5adf6b00005e7158ca493bb92477a762bf3ea9432/  in an other computer. But, the webrtc can not connect correct. **Because the android cuttlefish run in the docker with bridge network. So, the webrtc sould connect though a turn server.**

I modify the golang source code and config file like behand, it will work ok

https://github.com/google/cloud-android-orchestration/blob/main/scripts/docker/conf.toml
```
[WebRTC]
STUNServers = ["stun:10.10.13.101:3478"]
UseTurnServers = true

[WebRTC.TURNServers]
URLs = ["turn:10.10.13.101:3478"]
Username = "vtv"
Credential = "123456"
```

https://github.com/google/cloud-android-orchestration/blob/main/pkg/app/config/config.go
```
type WebRTCConfig struct {
	STUNServers []string
	UseTurnServers bool
	TURNServers *WebRTCTurnServersConfig
}

type WebRTCTurnServersConfig struct {
	URLs           []string
	Username       string
	Credential     interface{}
}
```

https://github.com/google/cloud-android-orchestration/blob/main/pkg/app/app.go
```
func NewApp(
	im instances.Manager,
	am accounts.Manager,
	oc *appOAuth2.Helper,
	es encryption.Service,
	dbs database.Service,
	webStaticFilesPath string,
	corsAllowedOrigins []string,
	webRTCConfig config.WebRTCConfig,
	config *config.Config) *App {
	return &App{im, am, oc, es, dbs, webStaticFilesPath, corsAllowedOrigins, buildInfraCfg(webRTCConfig), config}
}

func buildInfraCfg(buildInfraCfg config.WebRTCConfig) apiv1.InfraConfig {
	stunServers := buildInfraCfg.STUNServers
[...]

PR #395, opened 2025-01-14
Add TURN server support into Cloud Orchestrator

0405ysj:


api/v1/signalingserver.go
```diff
@@ -28,5 +28,8 @@ type InfraConfig struct {
 }
 
 type IceServer struct {
-	URLs []string `json:"urls"`
+	URLs               []string `json:"urls"`
+	Username           string   `json:"username,omitempty"`
```
ikicha, line 32: should the added fields(like Username, Credential) be an array as well as URLs?(not sure about IceTransportPolicy)

ikicha (changes requested):

api/v1/signalingserver.go
0405ysj: This struct is for single `IceServer`. In a single `IceServer`, there could be multiple URLs.

BTW, I think I did wrong on `IceTransportPolicy`. It shouldn't be in `IceServer`. On the other hand, it seems okay for not specifying `IceTransportPolicy` at the default configuration. I'll remove this from this PR.

ikicha: Ack, I thought we needs username/credential array, but https://w3c.github.io/webrtc-pc/#dom-rtciceserver says it is just string field, not array.

pkg/app/config/config.go
```diff
@@ -23,11 +23,13 @@ import (
 	"github.com/google/cloud-android-orchestration/pkg/app/instances"
 	"github.com/google/cloud-android-orchestration/pkg/app/secrets"
 
+	apiv1 "github.com/google/cloud-android-orchestration/api/v1"
 	toml "github.com/pelletier/go-toml"
 )
 
 type WebRTCConfig struct {
```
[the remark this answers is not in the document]
0405ysj: Do you want me to support `iceTransportPolicy` into TURN server configuration as well? I think we don't need this essentially. But I have no strong opinion on here.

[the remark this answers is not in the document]
ikicha: I think the icetransportpolicy is unlikely to be overridden.. let it be until there is a request about that.

ikicha (approved):
@ser-io 

Context: https://github.com/google/cloud-android-orchestration/issues/382 we want to set up turn server in CO as well as stun server is set

pkg/app/config/config.go
```diff
@@ -23,11 +23,13 @@ import (
...
 	STUNServers []string
+	TURNServers []apiv1.IceServer
```
ser-io, line 32: Can we just have one field: `Servers []apiv1.IceServer`?

ser-io (changes requested):

pkg/app/config/config.go
0405ysj: The only reason for introducing `TURNServers` was to keep compatibility with old `cvdr.toml`. Is it okay not considering compatibility?

ser-io (approved):

pkg/app/config/config.go
0405ysj: Oh, sorry. Yes, I meant `conf.toml`.  I'll be modifying on WebRTCConfig.

0405ysj: Done, I found there's no structural difference between `apiv1.InfraConfig` and `config.WebRTCConfig`, I removed that config.

PR #403, opened 2025-01-17
Support accelerator flag in cli and cvdr.toml

yippyyipyip:


pkg/cli/cli.go
```diff
@@ -82,7 +82,8 @@ const (
 const (
 	gcpMachineTypeFlag    = "gcp_machine_type"
 	gcpMinCPUPlatformFlag = "gcp_min_cpu_platform"
-	gcpBootDiskSizeGB     = "gcp_boot_disk_size_gb"
+	gcpBootDiskSizeGBFlag = "gcp_boot_disk_size_gb"
```
ser-io, line 85: Thanks for noticing `gcpBootDiskSizeGB` is missing the `Flag` suffix. However, let's have this fix in a different commit to keep this commit focused on accelerator config.

pkg/cli/cli.go
```diff
@@ -82,7 +82,8 @@ const (
...
+	acceleratorFlag       = "gcp_accelerator"
```
ser-io, line 86: let's name it: `gcpAcceleratorFlag`

pkg/cli/cli.go
```diff
@@ -541,6 +542,7 @@ func hostCommand(opts *subCommandOpts) *cobra.Command {
 }
 
 func cvdCommands(opts *subCommandOpts) []*cobra.Command {
+	acceleratorFlagValues := []string{}
```
ser-io, line 545: nit: `gcpAcceleratorFlagValues`

pkg/cli/cli.go
```diff
@@ -644,6 +651,10 @@ func cvdCommands(opts *subCommandOpts) []*cobra.Command {
 		create.Flags().StringVar(f.ValueRef, name, f.Default, f.Desc)
 		create.MarkFlagsMutuallyExclusive(hostFlag, name)
 	}
+	create.Flags().Int64Var(&createFlags.GCP.BootDiskSizeGB, "host_"+gcpBootDiskSizeGBFlag,
+		opts.InitialConfig.DefaultService().Host.GCP.BootDiskSizeGB, gcpBootDiskSizeGBDesc)
+	create.Flags().StringSliceVar(&acceleratorFlagValues, acceleratorFlag,
```
ser-io, line 656: let's rather add a new entry to the `createHostFlags` array above:

```
		{
			ValueRef: &gcpAcceleratorFlagValues,
			Name:     gcpAcceleratorFlag,
			Default:  opts.InitialConfig.DefaultService().Host.GCP.AcceleratorConfigs,
			Desc:     acceleratorFlagDesc,
		},
```

ser-io (changes requested):

pkg/cli/cli.go
yippyyipyip: I didn't add a new entry to createHostFlags because ValueRef has a type *string whereas gcpAcceleratorFlagValues is a string array. I suppose an option is to make the accelerator flag a string instead of string array even though the underlying accelerator config can accept multiple values?

ser-io: I see, let's keep it as is then. 

We still need to add the `host_` prefix to the flag name, similar to other flags in `createHostFlags`, it's flag affecting the host and not the virtual device.

ser-io (changes requested):

pkg/cli/cli.go
yippyyipyip: Is it intentional that the variable names `gcpMinCPUPlatformFlag` and `gcpMachineTypeFlag` are prefixed with 'host_' only for the create command but not the host command?

ser-io: yes. for the `host` command the prefix is redundant.

`cvdr create --host_gcp_accelerator` is equal to `host=$(cvdr host create --gcp_accelerator) && cvdr create --host=$host`

yippyyipyip: Fixed

ser-io (approved):

ser-io (changes requested):
Let's have only 1 commit.

Updating the `TestLoadFullConfig` should be part of the initial commit as well to avoid a misleading failure when performing a `git bisection`.

ser-io (approved):

yippyyipyip:
Fixed gofmt error

PR #406, opened 2025-01-31
Use full Spanner resource path in database service

yippyyipyip:


ser-io (approved):

cmd/cloud_orchestrator/main.go
```diff
@@ -144,7 +145,11 @@ func LoadDatabaseService(config *config.Config) database.Service {
 	case database.InMemoryDBType:
 		dbs = database.NewInMemoryDBService()
 	case database.SpannerDBType:
-		dbs = database.NewSpannerDBService(config.DatabaseService.Spanner.DatabaseName)
+		dbResource := fmt.Sprintf("projects/%s/instances/%s/databases/%s",
+			config.InstanceManager.GCP.ProjectID,
```
ser-io, line 149: @jemoreira should we fix the database project to the GCE project or we should be able to set another GCP project name?

ser-io (changes requested):

cmd/cloud_orchestrator/main.go
jemoreira: Being able to set a different one provides more flexibility IMHO.

yippyyipyip: Added a ProjectID field to the Spanner config

ser-io (approved):

PR #408, opened 2025-02-11
Sharing user artifact across host instances with dockerIM

0405ysj:
Corresponding PR for HO is https://github.com/google/android-cuttlefish/pull/882.

### DIFF
--- a/pkg/cli/cvd.go
+++ b/pkg/cli/cvd.go
@@ -16,13 +16,16 @@ package cli
 
 import (
 	"bytes"
+	"crypto/sha1"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"net/url"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sort"
 	"strings"
 	"text/template"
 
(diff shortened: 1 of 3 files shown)


pkg/cli/cvd.go
```diff
@@ -708,30 +699,56 @@ func envVar(name string) (string, error) {
 	return os.Getenv(name), nil
 }
 
-func uploadFiles(srv hoclient.HostOrchestratorService, uploadDir string, names []string, statePrinter *statePrinter) error {
+func getHashSHA1(files []string) (string, error) {
+	sort.Slice(files, func(x, y int) bool {
```
ikicha, line 703: https://pkg.go.dev/slices#SortFunc
it looks like slices.SortFunc is recommended.

BTW, why do we want to sort the list based on Base? is just slices.Sort enough?

In addition, we might want to sort based on fixed locale for stable result.. (https://pkg.go.dev/golang.org/x/text/collate#Collator.Sort)

0405ysj: > I see.. I think we'd better use filepath.Rel instead of using only file name.

Isn't `filepath.Rel` no-op?

> similar issue: b/148434609, the sort result can be different based on locale. maybe calling this API from different machine with different locale might cause unstable sort result. it might be overkill, so it's okay to just leave the comment as TODO or do it as follow-up items.

Isn't this because of the behavior of `find` command, not sorting?

ikicha (approved):

PR #421, opened 2025-03-17
Add timeout for Listen() while waiting for connection from ADB server.

k311093:
This prevents infinite waiting in the case that the ADB server does not respond to the ADB connect request.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -140,6 +140,10 @@ const (
 	iceConfigFlagDesc = "Path to file containing the ICE configuration to be used in the underlaying WebRTC connection"
 )
 
+const (
+	adbListenTimeoutSec = 30
+)
+
 type AsArgs interface {
 	AsArgs() []string
 }
@@ -1366,9 +1370,21 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
 			c.PrintErrf("Failed to connect ADB to device: %v\n", err)
 		}
 	}()
+
+	tcpListener, ok := l.(*net.TCPListener)
+	if !ok {
+		c.PrintErrf("Listener is not a TCPListener\n")
+		return errors.New("Listener is not a TCPListener")
+	}
+	tcpListener.SetDeadline(time.Now().Add(time.Duration(adbListenTimeoutSec) * time.Second))
+
 	tcpConn, err := l.Accept()
 	if err != nil {
-		c.PrintErrf("Failed to accept ADB socket: %v", err)
+		if ne, ok := err.(net.Error); ok && ne.Timeout() {
+			c.PrintErrf("Waiting connection from ADB server timed out: ")
+		} else {
+			c.PrintErrf("Failed to accept ADB socket: %v", err)
+		}
 		return err
 	}
 	defer tcpConn.Close()
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -1365,16 +1365,19 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
 		return err
 	}
 	adbPort := l.Addr().(*net.TCPAddr).Port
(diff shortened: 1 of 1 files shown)


pkg/cli/cli.go
```diff
@@ -1363,9 +1367,20 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
 			c.PrintErrf("Failed to connect ADB to device: %v\n", err)
```
0405ysj, line 1367: Can't we retrieve this error at the outside of goroutine via channel, and handle it there?

pkg/cli/cli.go
```diff
@@ -1363,9 +1367,20 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
...
 		}
 	}()
+
+	tcpListener, ok := l.(*net.TCPListener)
+	if !ok {
+		c.PrintErrf("Listener is not a TCPListener\n")
```
0405ysj, line 1373: Not only printing error message, shouldn't we return this function with an error?

0405ysj (changes requested):

pkg/cli/cli.go
k311093: Added return statement when type assertion is failed.

ikicha (approved):

pkg/cli/cli.go
k311093: Used channel to retrieve errors from ADB proxy connect goroutine.

Also changed to use goroutine and channel to retrieve errors and result from l.Accept() to handle errors in the same place.

0405ysj (approved):

pkg/cli/cli.go
```diff
@@ -1362,27 +1362,55 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
...
+	}(adbErrCh)
 
-	tcpListener, ok := l.(*net.TCPListener)
-	if !ok {
-		c.PrintErrf("Listener is not a TCPListener\n")
-		return errors.New("Listener is not a TCPListener")
-	}
-	tcpListener.SetDeadline(time.Now().Add(time.Duration(adbListenTimeoutSec) * time.Second))
+	listenResultCh := make(chan net.Conn)
+	listenErrCh := make(chan error)
```
jemoreira, line 1375: nit: No need for two channels if you create a small struct to hold the two values:
```
type listenResult struct {
    conn net.Conn
    err  error
}
listenCh := make(chan, listenResult)
```

pkg/cli/cli.go
```diff
@@ -1362,27 +1362,55 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
...
+		}
+		listenResultCh <- tcpConn
+	}(listenResultCh, listenErrCh)
+
+	var tcpConn net.Conn
+	for tcpConn == nil {
+		select {
+		case err := <-adbErrCh:
+			if err != nil {
+				return err
```
jemoreira, line 1402: nit: An early return here would cause the other channels to not be read from, which in turn would block the routine trying to send on that channel. The same happens with the return in line 1406. That's unlikely to cause major issues but it would be cleaner to make sure that both routines finish before returning from this function. That can be done by receiving from `adbErrCh` first and then a `select` with the other two channels (a `select` won't be needed if you follow my previous advice).
Or, just don't use a routine for accepting on the listener, that has the same effect as receiving from those channels before receiving from `adbErrCh`.

jemoreira (approved):

pkg/cli/cli.go
k311093: Merged separated channels into one channel.

pkg/cli/cli.go
k311093: Removed go routine for listener, and modified codes to read from adbErrCh channel to avoid the situation mentioned.

PR #422, opened 2025-03-17
Service flag should be passed in ServiceFlags.AsArgs()

0405ysj:


k311093 (approved):

ikicha (commented):
can you share the context of this?

the context of f.Service becomes f.ServiceURL and Zone and Proxy via `func (f *ServiceFlags) Update(s *Service)`, isn't it?

0405ysj:
In my investigation, `ServiceURL`, `Zone`, and `Proxy` wasn't sufficient not to pass `Service`.

Problem occurs when:
1. Trying `cvdr connect` with non-default service.
2. `Service` flag is not passed into agent command like `websocket`.
3. Agent command tries to read auth configs with `Service` flag. [link](https://github.com/google/cloud-android-orchestration/blob/a344e3e492644f0ea4748a21d8c3a2ed03fb18f0/pkg/cli/cli.go#L1634-L1667)

PR #423, opened 2025-04-02
Add reconnection logic when connection is closed at ADB WebSocket agent

k311093:
For supporting temporary connection closure while executing "adb root" or "adb unroot"

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"math"
 	"net"
 	"net/http"
 	"net/url"
@@ -28,6 +29,7 @@ import (
 	"os/user"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	client "github.com/google/cloud-android-orchestration/pkg/client"
@@ -1328,34 +1330,6 @@ func runConnectionWebrtcAgentCommand(flags *ConnectFlags, c *command, args []str
 	return nil
 }
 
-func connectADBWebSocketDirect(serviceURL string, device string) (*websocket.Conn, error) {
-	// Connect to ADB proxy WebSocket directly by using serviceURL as Operator URL.
-	//   wss://127.0.0.1:1443/devices/cvd-1/adb
-	url, err := url.Parse(serviceURL)
-	if err != nil {
-		return nil, fmt.Errorf("failed to parse URL %s: %w", serviceURL, err)
-	}
-	switch p := &url.Scheme; *p {
-	case "https":
-		*p = "wss"
-	case "http":
-		*p = "ws"
-	default:
-		return nil, fmt.Errorf("unknown scheme %s", *p)
-	}
-	url = url.JoinPath("devices", device, "adb")
-	dialer := websocket.Dialer{
-		TLSClientConfig: &tls.Config{
-			InsecureSkipVerify: true,
-		},
-	}
-	wsConn, _, err := dialer.Dial(url.String(), nil)
-	if err != nil {
-		return nil, fmt.Errorf("failed to connect WebSocket %s: %w", url.String(), err)
-	}
-	return wsConn, nil
-}
-
 func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []string, opts *subCommandOpts) error {
 	// Open local TCP port that ADB server connects to and make ADB server connect
 	// to the port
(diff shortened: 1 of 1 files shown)


pkg/cli/cli.go
```diff
@@ -1445,14 +1456,42 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
...
 
-	// Redirect WebSocket to ADB TCP port
-	wsWrapper := newWsIoWrapper(c, wsConn)
-	go func() {
-		io.Copy(wsWrapper, tcpConn)
-		wsWrapper.Close()
-	}()
-	io.Copy(tcpConn, wsWrapper)
-	return nil
+	adbConnectRetryDelaySecs := []int{0, 1, 1, 1, 1, 1, 5, 10, 10}
```
0405ysj, line 1459: Rather than having an array of list, could we apply any simple formula? For example, getting twice larger than previous retry delay.

pkg/cli/cli.go
```diff
@@ -1445,14 +1456,42 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
...
+
+	for {
+		// Redirect WebSocket to ADB TCP port
+		wsWrapper := newWsIoWrapper(c, wsConn)
+		go func() {
+			io.Copy(wsWrapper, tcpConn)
+			wsWrapper.Close()
+		}()
+		io.Copy(tcpConn, wsWrapper)
+		tcpConn.Close()
```
0405ysj, line 1469: QQ: Is the precondition for invoking `tcpConn.Close()` when the connection is lost?

pkg/cli/cli.go
```diff
@@ -1445,14 +1456,42 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
...
+
+		// Try to re-establish proxy when connection is closed
+		// For the cases that the connection closes temporary (e.g. "abd root")
+		tcpListener.SetDeadline(time.Now().Add(time.Duration(adbListenTimeoutSec) * time.Second))
```
0405ysj, line 1473: Why don't we construct a function? I can see dup'd line above.

pkg/cli/cli.go
```diff
@@ -1445,14 +1456,42 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
...
+		tcpConn, err = l.Accept()
+		if err != nil {
+			c.PrintErrf("Failed to accept ADB socket while reconnecting: %v", err)
+			return err
+		}
+
+		retryCount := 0
+
+		for retryCount = 0; retryCount < len(adbConnectRetryDelaySecs); retryCount++ {
```
0405ysj, line 1482: You may want to use either
```
var retryCount Int
for retryCount = 0; retryCount < len(adbConnectRetryDelaySecs); retryCount++ {
```
or 
```
retryCount := 0
for ; retryCount < len(adbConnectRetryDelaySecs); retryCount++ {
```
?

0405ysj (changes requested):

jemoreira (commented):
I left comments about some individual problems I found, but I think all of them could be addressed by changing the structure of the agent function to something like this:
```
  func runConnectionWebSocketAgentCommand(...) {
  »   l := ... // tcp listener created here
  »   for {
  »   »   if err := ConnectAndForwardOnce(l, ...); err != nil {
  »   »   »   // Don't even try again, something went wrong
  »   »   »   return err
  »   »   }
  »   »   // no error means connection closed normally, try again.
  »   }
  }
  
  func ConnectAndForwardOnce(l, ...) error {
  »   tcpConn, connErr := acceptADBConnection(l, ...)
  »   if connErr != nil {
  »   »   return connErr
  »   }
  »   // Set this up immediately after the connection is known to be open
  »   defer tcpConn.Close()
  
  »   devConn, devErr := connectToDeviceADB(...)
  »   if devErr != nil {
  »   »   return devErr
  »   }
  
  »   forwardCh
  »   go func(...) {
  »   »   defer devConn.Close()
  »   »   io.Copy(devConn, tcpConn)
  »   }(forwardCh)
  
  »   io.Copy(tcpConn, devConn)
  »   // Note that the tcpConn is closed here and the devConn is closed in the routine, this is important to avoid SIGPIPE
  »   return nil
  }
  
  func acceptADBConnection(l, ...) (tcpConn, error) {
  »   // Ask ADB to connect to our listener in a different routine so we can call accept()
  »   adbCh := ConnectADB(l.Port)
  
  »   l.SetDealine(...)
  »   tcpConn, connErr := l.Accept()
  
  »   // Make sure to read the adb connection result
  »   if adbErr := <- adbCh; adbErr != nil
  »   »   // Just log it, if this failed the accept timeout, if the accept succeeded we don't really care about this.
  »   }
  
  »   if connErr != nil {
  »   »   return nil, err
  »   }
  
  »   return tcpConn, nil
  }
  
  func connectToDeviceADB(...) {
  »   // This implementation differs between websockets through HO, websockets direct
  »   // and webrtc. Ideally these should be implemented in a way that the rest of the
  »   // code is the same and can be reused without duplication. The one issue I see
  »   // is the websocket connection not implementing io.Read/Write (needed for io.Copy)
  »   // in which case we could wrap that connection in a struct that implements it.
  }

```

pkg/cli/cli.go
k311093: Changed to use math.Power to get delay time as exponential. But used int(n / 2) as exponent to prevent delay time grow too fast.

pkg/cli/cli.go
k311093: For tcpConn.Close(), this is executen when io.Copy(tcpConn, wsWrapper) is returned which indicates that wsWrapper is closed.

pkg/cli/cli.go
k311093: Setting timeout and listen is merged into one function while changing structure of the code.

pkg/cli/cli.go
k311093: Changed to use first suggestion,

pkg/cli/cli.go
```diff
@@ -1396,28 +1422,13 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
...
-		}
-		wsConn, err = srvClient.HostService(flags.host).ConnectADBWebSocket(device)
-		if err != nil {
-			return err
-		}
+	device := args[0]
+	wsConn, err = connectADBWebSocket(flags, c, opts, device)
+	if err != nil {
+		c.PrintErrf("Connect to ADB websocket failed: ")
+		return err
```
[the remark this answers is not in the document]
k311093: Handle this case by using `defer tcpConn.Close()`

pkg/cli/cli.go
```diff
@@ -1445,14 +1456,42 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
...
-	wsWrapper := newWsIoWrapper(c, wsConn)
-	go func() {
-		io.Copy(wsWrapper, tcpConn)
-		wsWrapper.Close()
-	}()
-	io.Copy(tcpConn, wsWrapper)
-	return nil
+	adbConnectRetryDelaySecs := []int{0, 1, 1, 1, 1, 1, 5, 10, 10}
+
+	for {
```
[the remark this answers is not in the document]
k311093: Changed to do things at the loop by applying the structure change at the other comment.

pkg/cli/cli.go
```diff
@@ -1364,95 +1338,159 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
...
-	}(adbErrCh)
+		c.PrintErrf("Failed to connect ADB to device: %v, retrying (%d / %d)\n", devErr, retryCount+1, maxRetryCount)
+		time.Sleep(time.Duration(math.Pow(2, float64(retryCount/2))) * time.Second)
+	}
+	if retryCount == maxRetryCount {
+		return devErr
+	}
+
+	if sendResultErr := sendConnectResult(c, adbPort); sendResultErr != nil {
+		return sendResultErr
```
0405ysj, line 1383: Does here need to close `devConn` before terminating `connectAndForwardOnce`?

pkg/cli/cli.go
```diff
@@ -1364,95 +1338,159 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
...
-			State: StateAsStr(FwdConnected),
-		},
-	}
-	output, err := json.Marshal(result)
-	if err != nil {
-		return err
-	} else {
-		c.Println(string(output))
-	}
+var sendResultOnce sync.Once
```
0405ysj, line 1462: Why do we need this?

pkg/cli/cli.go
```diff
@@ -1364,95 +1338,159 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
...
 		return err
 	}
+	device := args[0]
 	adbPort := l.Addr().(*net.TCPAddr).Port
-	adbErrCh := make(chan error)
-	go func(adbErrCh chan error) {
-		err := opts.ADBServerProxy.Connect(adbPort)
-		if err != nil {
-			c.PrintErrf("Failed to connect ADB to device: %v\n", err)
+	err = opts.ADBServerProxy.Connect(adbPort)
```
0405ysj, line 1343: OOC, don't we need channel based solution anymore?

k311093: Yes, I checked the code without channel based solution and it worked fine. I suspected `opts.ADBServerProxy.Connect` is blocking before, but found that it is not blocking and does not needs to be in the goroutine.

0405ysj: Acknowledged. Thanks for your verification!

pkg/cli/cli.go
k311093: sendResultOnce is for confirming `sendConnectResult` function run just once (send result to parent and then daemonize is needed be executed just once)

pkg/cli/cli.go
k311093: Added `devConn.Close()` when error is happened except max retry count is reached while trying to connect.

0405ysj (approved):

pkg/cli/cli.go
```diff
@@ -1364,95 +1338,160 @@ func runConnectionWebSocketAgentCommand(flags *ConnectFlags, c *command, args []
...
+	}
+
+	forwardCh := make(chan interface{})
+	go func(forwardCh chan interface{}) {
+		io.Copy(devConn, tcpConn)
+		devConn.Close()
+		forwardCh <- nil
+	}(forwardCh)
+	io.Copy(tcpConn, devConn)
+	tcpConn.Close()
```
jemoreira, line 1394: not needed with the `defer` on line 1362.

jemoreira (approved):

jemoreira:
> @jemoreira Hi, can you please review this PR? I've updated the comments from the previous review.
> 
> Thanks!

Sorry for the delay, was waiting for a response on that remaining unresolved conversation. I see that you already did it so I approved the PR.

pkg/cli/cli.go
k311093: I used tcpConn.Close to interrupt io.Copy(...) in the goroutine above because in the main thread, it waits forwardCh and tio.Copy(...) function in the goroutine needs to be returned for avoiding blocking.

I changed `tcpConn.Close()` to `tcpConn.SetDeadline(time.Now())` to interrupt `io.Copy...` and avoid double closure.

k311093:
> > @jemoreira Hi, can you please review this PR? I've updated the comments from the previous review.
> > Thanks!
> 
> Sorry for the delay, was waiting for a response on that remaining unresolved conversation. I see that you already did it so I approved the PR.

I didn't knew that there was unresolved conversation. Sorry for the confusing.

jemoreira:
> > > @jemoreira Hi, can you please review this PR? I've updated the comments from the previous review.
> > > Thanks!
> > 
> > 
> > Sorry for the delay, was waiting for a response on that remaining unresolved conversation. I see that you already did it so I approved the PR.
> 
> I didn't knew that there was unresolved conversation. Sorry for the confusing.

Just realized I didn't leave that comment as part of a review so it couldn't be "resolved". My bad.

pkg/cli/cli.go
jemoreira: Good point, didn't notice that deadlock.

Could you leave a comment above the `SetDeadline` call explaining that it's intended to unblock the goroutine?

k311093:
> > > > @jemoreira Hi, can you please review this PR? I've updated the comments from the previous review.
> > > > Thanks!
> > > 
> > > 
> > > Sorry for the delay, was waiting for a response on that remaining unresolved conversation. I see that you already did it so I approved the PR.
> > 
> > 
> > I didn't knew that there was unresolved conversation. Sorry for the confusing.
> 
> Just realized I didn't leave that comment as part of a review so it couldn't be "resolved". My bad.

No problem. Thanks! :)

pkg/cli/cli.go
k311093: Added comment about that.

PR #425, opened 2025-04-15
DockerIM can select updating CF deb packages during host creation or not

0405ysj:
Bug: b/408156615

Since we've observed updating CF debian packages at the entrypoint of docker instance can reach to the timeout of host creation, this change is for providing option that running docker image without performing `apt update` & `apt install`.

### DIFF
--- a/pkg/app/instances/docker.go
+++ b/pkg/app/instances/docker.go
@@ -38,13 +38,15 @@ const DockerIMType IMType = "docker"
 
 type DockerIMConfig struct {
 	DockerImageName      string
+	UpdateDebPackages    bool
 	HostOrchestratorPort int
 }
 
 const (
-	dockerLabelCreatedBy      = "created_by"
-	dockerLabelKeyManagedBy   = "managed_by"
-	dockerLabelValueManagedBy = "cloud_orchestrator"
+	dockerLabelCreatedBy           = "created_by"
+	dockerLabelKeyManagedBy        = "managed_by"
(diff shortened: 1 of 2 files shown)


0405ysj:
@ikicha Could you review this PR?

ser-io:
Why are the cuttlefish debian packages getting update at host creation? The're should be a controlled way to roll out and roll back new version of cuttlefish packages to your host fleet, this is achievable by updating or reverting the "cuttlefish-orchestration:latest"  used in your deployment. Users of the orchestration solution do not control when new packages are pushed into the public codelinaro repository.

With this patch, you're making auto-update configurable defaulting to false, hence no auto-update. I suggest to remove the auto-update part from the entrypoint and avoid having this hiding logic.

0405ysj:
You're right. I was focusing on supporting non-auto-update case again, but now I have a doubt again about whether we still need to support auto-update feature or not.

Reasons for supporting auto-update was to provide latest CF debian packages though the latest release of docker image doesn't have them, due to the release frequency difference between CF debian packages and docker images. And also we assumed there won't be lots of case requiring non-auto-update case.

But today I think we lost all justifications for keeping auto-update feature.
@ikicha HDYT? I'd like Sergio's opinion, I want to get rid of it either.

ikicha:
I agree with that, we can remove auto-update feature inside docker.

0405ysj:
Okay. Let's close this PR, and proceed with getting rid of auto-updating CF debian package inside docker image.

PR #441, opened 2025-07-16
Presubmit check executes cvdr E2E tests with CO(on-premise configuration)

0405ysj:
This PR is for enabling presubmit check of cvdr E2E test based on CO with on-premise configuration.

This PR brought artifacts from:
- `cuttlefish-orchestration`: Pull docker image from Artifact Registry.
- `cuttlefish-cloud-orchestrator`: Load docker image uploaded by previous workflow job.
- `cuttlefish-cvdremote`: Install debian package uploaded by previous workflow job.


.github/workflows/presubmit.yaml
```diff
@@ -0,0 +1,123 @@
+name: Presubmit
+
+on:
+  pull_request:
+  push:
+    branches-ignore:
+      - main  # push events to main branch occur after PRs are merged, when the same checks were run
+
+jobs:
+  build-test:
```
jemoreira, line 10: why do you need to merge all these into the same file?

.github/workflows/presubmit.yaml
```diff
@@ -0,0 +1,123 @@
+name: Presubmit
+
+on:
+  pull_request:
+  push:
+    branches-ignore:
+      - main  # push events to main branch occur after PRs are merged, when the same checks were run
```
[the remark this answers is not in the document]
0405ysj: That should be okay. This Github repository is [using](https://github.com/google/cloud-android-orchestration/actions/workflows/build_and_test.yaml?query=event%3Apush) [merge queue](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#how-merge-queues-work). In my understanding, a PR is validated with previous PRs enrolled into the merge queue.

.github/workflows/presubmit.yaml
0405ysj: I have 2 reasons to prefer merge `build-cuttlefish-cloud-orchestrator-x86_64-docker-image` here.
- The job building x86_64 docker image containing CO should depend on the job performing e2e test of `cvdr` with on-premise config.
- When the job is running on the different workflow, `action/download-artifact` should pass ID of the workflow.

On the other hand, I believe above reasons can increase code complexity but not a technical blocker. ([on.workflow_call.outputs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-of-onworkflow_calloutputs), [jobs.<job_id>.steps[*].uses](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-an-action-inside-a-different-private-repository-than-the-workflow))

I speculate you prefer to leave jobs in the separate workflows, could I hear the reason why it can be better? Depending on your answer, I may refactor Github workflow `.yaml` files since I want to download artifact from `build-cuttlefish-cvdremote-debian-package` either.

jemoreira: It's just that smaller files are easier to read and maintain. We can split it later if it becomes too big.

jemoreira (approved):

PR #457, opened 2025-08-12
Introduce a shell script for creating GCE x86_64 image which can run CO

0405ysj:


scripts/on-premises/gcp/gce_bare_metal_image_builder.sh
```diff
@@ -0,0 +1,166 @@
...
+RANDOM_SUFFIX=$(cat /dev/urandom | tr -dc 'a-z0-9' | head -c 8)
+IMAGE_BUILDER_INSTANCE=cf-image-builder-$RANDOM_SUFFIX
+IMAGE_NAME=cf-bare-metal-$RANDOM_SUFFIX
+
+create_image_builder_instance() {
+    if [[ $ARCH == "amd64" ]]; then
+        $GCLOUD_COMMAND --project $PROJECT \
+            compute instances create $IMAGE_BUILDER_INSTANCE \
+            --image-project=debian-cloud \
+            --image-family=debian-12 \
```
ikicha, line 88: is it still using debian-12? As I barely remember, we might want to use trixie instead? @adelva1984

scripts/on-premises/gcp/gce_bare_metal_image_builder.sh
```diff
@@ -0,0 +1,166 @@
...
+        echo 'GCE instance is not ready yet, waiting...'
+        sleep 5
+    done
+    exec_command_on_image_builder_instance "
+        echo \"deb http://deb.debian.org/debian bookworm-backports main\" |\
+            sudo tee /etc/apt/sources.list &&
+        sudo apt update &&
+        sudo apt install -y linux-base -t bookworm-backports &&
+        sudo apt install -y linux-image-cloud-$ARCH -t bookworm-backports &&
+        sudo apt install -y docker.io &&
```
ikicha, line 130: per docker's official guide, they don't recommend docker.io package.. https://docs.docker.com/engine/install/debian/

scripts/on-premises/gcp/gce_bare_metal_image_builder.sh
0405ysj: Since we don't have `debian-13` now AFAIK. I can find Trixie is released a couple of days ago, but I haven't found anything about Debian 13 on GCP yet. https://www.debian.org/News/2025/20250809

scripts/on-premises/gcp/gce_bare_metal_image_builder.sh
0405ysj: Acknowledged. I would try to switch into docker-ce. Thanks for letting me know.

0405ysj: Done.

scripts/on-premises/gcp/gce_image_builder.sh
```diff
@@ -0,0 +1,173 @@
...
+        sudo docker pull \
+            $DOWNLOAD_LOCATION-docker.pkg.dev/android-cuttlefish-artifacts/cuttlefish-orchestration/cuttlefish-orchestration &&
+        sudo mkdir -p /etc/cloud_orchestrator &&
+        sudo wget -O /etc/cloud_orchestrator/conf.toml \
+            https://artifactregistry.googleapis.com/download/v1/projects/android-cuttlefish-artifacts/locations/$DOWNLOAD_LOCATION/repositories/cloud-orchestrator-config/files/on-premise-single-server:main:conf.toml:download?alt=media &&
+        sudo docker run --restart unless-stopped -d -p 8080:8080 -e CONFIG_FILE="/conf.toml" \
+            -v /etc/cloud_orchestrator/conf.toml:/conf.toml \
+            -v /var/run/docker.sock:/var/run/docker.sock \
+            -t $DOWNLOAD_LOCATION-docker.pkg.dev/android-cuttlefish-artifacts/cuttlefish-orchestration/cuttlefish-cloud-orchestrator:latest
+    "
```
jemoreira, line 149: Why does the cloud orchestrator need to be started on the image builder?

Shouldn't it be stopped before shutting it down and creating an image from it?

scripts/on-premises/gcp/gce_image_builder.sh
```diff
@@ -0,0 +1,173 @@
+#!/bin/bash
+
+set -e
+
+DEFAULT_ZONE=us-central1-a
+DEFUULT_ARCH=amd64
```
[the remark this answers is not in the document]
0405ysj: Thanks, I missed it. Done.

scripts/on-premises/gcp/gce_image_builder.sh
```diff
@@ -0,0 +1,173 @@
...
+                ;;
+            -c|--gcloud_command)
+                if [ -z "$GCLOUD_COMMAND" ]; then
+                    echo "Missing gcloud command"
+                    exit 1
+                fi
+                GCLOUD_COMMAND="$2"
+                shift 2
+                ;;
+            --)
```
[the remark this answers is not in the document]
0405ysj: Done.

scripts/on-premises/gcp/gce_image_builder.sh
0405ysj: > Shouldn't it be stopped before shutting it down and creating an image from it?

No, it was okay. I checked it kept running in the GCE instance with the image.

> Why does the cloud orchestrator need to be started on the image builder?

The only purpose is to ensure CO is running on GCE instance, so it doesn't need to be. But I didn't have any reasons to choose alternatives since docker itself supports such option.

scripts/on-premises/gcp/gce_image_builder.sh
```diff
@@ -0,0 +1,168 @@
+#!/bin/bash
```
ser-io, line 1: Please create rather a Golang command under https://github.com/google/android-cuttlefish/tree/main/tools/baseimage/cmd.

The original cuttlefish script to create host images was written in bash and had to be rewritten to Golang https://cs.android.com/android/platform/superproject/main/+/main:device/google/cuttlefish/tools/create_base_image.go. 

Existing command https://github.com/google/android-cuttlefish/blob/main/tools/baseimage/cmd/create_gce_x86_64_image/main.go has the GCE API calls you need to create the image.

scripts/on-premises/gcp/gce_image_builder.sh
```diff
@@ -0,0 +1,168 @@
...
+}
+
+stop_image_builder_instance() {
+    $GCLOUD_COMMAND --project $PROJECT \
+        compute instances stop $IMAGE_BUILDER_INSTANCE \
+        --zone=$ZONE
+}
+
+create_image() {
+    $GCLOUD_COMMAND --project $PROJECT \
```
ser-io, line 154: From my reading it seems you're creating the image out of the boot disk rather than creating the image out of an attached disk.

Creating the image out of the boot disk is problematic as you'll be porting GCE startup information to your image. Alternative is to build the image out of an attached disk. See https://github.com/google/android-cuttlefish/blob/main/tools/baseimage/cmd/create_gce_x86_64_image/main.go for more info,

ser-io (changes requested):

scripts/on-premises/gcp/gce_bare_metal_image_builder.sh
ser-io: Yes, we want to migrate the scripts to debian-13, however these scripts are based on GCP's Debian images, even though debian 13 was officially released, GCP hasn't rolled out a debian-13 image yet.

scripts/on-premises/gcp/gce_image_builder.sh
jemoreira: The way it seems to be working is that it's killed on VM shutdown and when it restarts (on a different VM with the same image) docker sees that it wasn't explicitly stopped and restarts it. While that might work most of the time, it could go wrong at some point, especially if the CO is caught writing any files or something like that. I'd prefer to do it in a different way, possibly by writing an init.d entry.

0405ysj:
Closing this PR as we've decided to move this logic into google/android-cuttlefish.

PR #462, opened 2025-08-21
Don't use statePrinter inside goroutines

jemoreira:
Because it's not thread safe. Removed the locking logic since it can't be used in multithreaded code anyways.


0405ysj:
Thanks for fixing it!

PR #521, opened 2026-04-15
Move waiting logic for host orchestrator settlement to backend

k311093:
While interacting with host orchestrator right after creating host
API is returned, there is a possibility of getting 503 service
temporarily unavailable error because host it created and booted but
host orchestrator is not executed / ready yet.

Handling logic for waiting host orchestrator readiness is currently
in the client logic. Moving this logic into cloud orchestrator API
endpoint would make the client that uses host chestrator API simpler.

Context: b/501288123

### DIFF
--- a/pkg/app/instances/docker.go
+++ b/pkg/app/instances/docker.go
@@ -188,6 +188,13 @@ func (m *DockerInstanceManager) waitCreateHostOperation(host string) (*apiv1.Hos
 				return nil, fmt.Errorf("failed to inspect docker container: %w", err)
 			}
 			if res.State.Running {
+				client, err := m.GetHostClient("local", host)
+				if err != nil {
+					return nil, err
+				}
+				if err := client.WaitForHostReady(); err != nil {
+					return nil, err
+				}
 				return &apiv1.HostInstance{
 					Name: host,
 				}, nil
--- a/pkg/app/instances/docker.go
+++ b/pkg/app/instances/docker.go
@@ -193,10 +193,7 @@ func (m *DockerInstanceManager) waitCreateHostOperation(host string) (*apiv1.Hos
 					return nil, err
 				}
 				// There is a delay between host creation and its readiness, wait for host readiness and return when it is ready.
-				readinessChecker, ok := client.(HostReadinessChecker)
-				if !ok {
-					readinessChecker = &BasicHostReadinessChecker{Client: client}
-				}
+				readinessChecker := &BasicHostReadinessChecker{Client: client}
 				if err := WaitForHostReady(readinessChecker, 2*time.Minute, 5*time.Second); err != nil {
 					return nil, err
 				}
--- a/pkg/app/instances/docker.go
+++ b/pkg/app/instances/docker.go
@@ -188,6 +188,18 @@ func (m *DockerInstanceManager) waitCreateHostOperation(host string) (*apiv1.Hos
 				return nil, fmt.Errorf("failed to inspect docker container: %w", err)
 			}
 			if res.State.Running {
+				client, err := m.GetHostClient("local", host)
+				if err != nil {
+					return nil, err
+				}
+				// There is a delay between host creation and its readiness, wait for host readiness and return when it is ready.
+				readinessChecker, ok := client.(HostReadinessChecker)
+				if !ok {
+					readinessChecker = &BasicHostReadinessChecker{Client: client}
+				}
+				if err := WaitForHostReady(readinessChecker, 2*time.Minute, 5*time.Second); err != nil {
+					return nil, err
+				}
 				return &apiv1.HostInstance{
 					Name: host,
 				}, nil
--- a/pkg/app/instances/gce.go
+++ b/pkg/app/instances/gce.go
@@ -78,28 +78,40 @@ func (m *GCEInstanceManager) ListZones() (*apiv1.ListZonesResponse, error) {
 	}, nil
 }
 
-func (m *GCEInstanceManager) GetHostAddr(zone string, host string) (string, error) {
-	instance, err := m.getHostInstance(zone, host)
-	if err != nil {
-		return "", err
-	}
-	ilen := len(instance.NetworkInterfaces)
+func getHostAddrWithIns(ins *compute.Instance) (string, error) {
+	ilen := len(ins.NetworkInterfaces)
 	if ilen == 0 {
-		log.Printf("host instance %s in zone %s is missing a network interface", host, zone)
+		log.Printf("host instance %s in zone %s is missing a network interface", ins.Name, ins.Zone)
 		return "", errors.NewInternalError("host instance missing a network interface", nil)
 	}
 	if ilen > 1 {
-		log.Printf("host instance %s in zone %s has %d network interfaces", host, zone, ilen)
+		log.Printf("host instance %s in zone %s has %d network interfaces", ins.Name, ins.Zone, ilen)
 	}
-	return instance.NetworkInterfaces[0].NetworkIP, nil
+	return ins.NetworkInterfaces[0].NetworkIP, nil
 }
 
-func (m *GCEInstanceManager) GetHostURL(zone string, host string) (*url.URL, error) {
-	addr, err := m.GetHostAddr(zone, host)
+func (m *GCEInstanceManager) GetHostAddr(zone string, host string) (string, error) {
+	ins, err := m.getHostInstance(zone, host)
+	if err != nil {
+		return "", err
+	}
+	return getHostAddrWithIns(ins)
+}
+
+func getHostURLWithIns(ins *compute.Instance, config *Config) (*url.URL, error) {
+	addr, err := getHostAddrWithIns(ins)
 	if err != nil {
 		return nil, err
 	}
-	return url.Parse(fmt.Sprintf("%s://%s:%d", m.Config.HostOrchestratorProtocol, addr, m.Config.GCP.HostOrchestratorPort))
+	return url.Parse(fmt.Sprintf("%s://%s:%d", config.HostOrchestratorProtocol, addr, config.GCP.HostOrchestratorPort))
+}
+
+func (m *GCEInstanceManager) GetHostURL(zone string, host string) (*url.URL, error) {
+	ins, err := m.getHostInstance(zone, host)
+	if err != nil {
+		return nil, err
+	}
+	return getHostURLWithIns(ins, &m.Config)
 }
 
 const operationStatusDone = "DONE"
@@ -253,16 +265,28 @@ func (m *GCEInstanceManager) WaitOperation(zone string, user accounts.User, name
 	if op.Status != operationStatusDone {
 		return nil, errors.NewServiceUnavailableError("Wait for operation timed out", nil)
 	}
-	getter := opResultGetter{Service: m.Service, Op: op}
+	getter := opResultGetter{
+		Service: m.Service,
+		Op:      op,
+		Config:  &m.Config,
+	}
 	return getter.Get()
 }
 
-func (m *GCEInstanceManager) GetHostClient(zone string, host string) (HostClient, error) {
-	url, err := m.GetHostURL(zone, host)
+func getHostClientWithIns(ins *compute.Instance, config *Config) (HostClient, error) {
+	url, err := getHostURLWithIns(ins, config)
 	if err != nil {
 		return nil, err
 	}
-	return NewNetHostClient(url, m.Config.AllowSelfSignedHostSSLCertificate), nil
+	return NewNetHostClient(url, config.AllowSelfSignedHostSSLCertificate), nil
+}
+
+func (m *GCEInstanceManager) GetHostClient(zone string, host string) (HostClient, error) {
+	ins, err := m.getHostInstance(zone, host)
+	if err != nil {
+		return nil, err
+	}
+	return getHostClientWithIns(ins, &m.Config)
 }
 
 func (m *GCEInstanceManager) getHostInstance(zone string, host string) (*compute.Instance, error) {
@@ -326,6 +350,7 @@ var (
 type opResultGetter struct {
 	Service *compute.Service
 	Op      *compute.Operation
+	Config  *Config
 }
 
 func (g *opResultGetter) Get() (any, error) {
@@ -362,7 +387,19 @@ func (g *opResultGetter) buildCreateInstanceResult() (*apiv1.HostInstance, error
 	if err != nil {
 		return nil, toAppError(err)
 	}
-	return BuildHostInstance(ins)
+
+	host, err := BuildHostInstance(ins)
+	if err != nil {
+		return nil, err
+	}
+	client, err := getHostClientWithIns(ins, g.Config)
+	if err != nil {
+		return nil, err
+	}
+	if err := client.WaitForHostReady(); err != nil {
+		return nil, err
+	}
+	return host, nil
 }
 
 // Converts compute API errors to AppError if relevant, return the same error otherwise
--- a/pkg/app/instances/gce.go
+++ b/pkg/app/instances/gce.go
@@ -18,9 +18,11 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"net/http"
 	"net/url"
 	"path"
 	"regexp"
+	"time"
 
 	apiv1 "github.com/google/cloud-android-orchestration/api/v1"
 	"github.com/google/cloud-android-orchestration/pkg/app/accounts"
@@ -79,27 +81,19 @@ func (m *GCEInstanceManager) ListZones() (*apiv1.ListZonesResponse, error) {
 }
 
 func (m *GCEInstanceManager) GetHostAddr(zone string, host string) (string, error) {
-	instance, err := m.getHostInstance(zone, host)
+	ins, err := m.getHostInstance(zone, host)
 	if err != nil {
 		return "", err
 	}
-	ilen := len(instance.NetworkInterfaces)
-	if ilen == 0 {
-		log.Printf("host instance %s in zone %s is missing a network interface", host, zone)
-		return "", errors.NewInternalError("host instance missing a network interface", nil)
-	}
-	if ilen > 1 {
-		log.Printf("host instance %s in zone %s has %d network interfaces", host, zone, ilen)
-	}
-	return instance.NetworkInterfaces[0].NetworkIP, nil
+	return m.getHostAddrWithIns(ins)
 }
 
 func (m *GCEInstanceManager) GetHostURL(zone string, host string) (*url.URL, error) {
-	addr, err := m.GetHostAddr(zone, host)
+	ins, err := m.getHostInstance(zone, host)
 	if err != nil {
 		return nil, err
 	}
-	return url.Parse(fmt.Sprintf("%s://%s:%d", m.Config.HostOrchestratorProtocol, addr, m.Config.GCP.HostOrchestratorPort))
+	return m.getHostURLWithIns(ins)
 }
 
 const operationStatusDone = "DONE"
@@ -253,16 +247,103 @@ func (m *GCEInstanceManager) WaitOperation(zone string, user accounts.User, name
 	if op.Status != operationStatusDone {
 		return nil, errors.NewServiceUnavailableError("Wait for operation timed out", nil)
 	}
-	getter := opResultGetter{Service: m.Service, Op: op}
-	return getter.Get()
+	getter := opResultGetter{
+		Service: m.Service,
+		Op:      op,
+		Config:  &m.Config,
+	}
+	res, err := getter.Get()
+	if err != nil {
+		return nil, err
+	}
+	if hostInst, ok := res.(*apiv1.HostInstance); ok && op.OperationType == "insert" {
+		return m.waitHostAvailability(zone, user, hostInst.Name)
+	}
+	return res, nil
 }
 
-func (m *GCEInstanceManager) GetHostClient(zone string, host string) (HostClient, error) {
-	url, err := m.GetHostURL(zone, host)
+func (m *GCEInstanceManager) waitHostAvailability(zone string, user accounts.User, host string) (*apiv1.HostInstance, error) {
+	ins, err := m.getHostInstance(zone, host)
+	if err != nil {
+		return nil, err
+	}
+	hostInstance, err := BuildHostInstance(ins)
+	if err != nil {
+		return nil, err
+	}
+	client, err := m.getHostClientWithIns(ins)
+	if err != nil {
+		return nil, err
+	}
+	if err := m.waitForOrchestrator(zone, hostInstance, client); err != nil {
+		return nil, err
+	}
+	return hostInstance, nil
+}
+
+type gceHostReadinessChecker struct {
+	basicChecker *BasicHostReadinessChecker
+	manager      *GCEInstanceManager
+	zone         string
+	hostName     string
+}
+
+func (g *gceHostReadinessChecker) IsHostReady() (bool, error) {
+	_, err := g.manager.Service.Instances.Get(g.manager.Config.GCP.ProjectID, g.zone, g.hostName).Context(context.TODO()).Do()
+	if err != nil {
+		if apiErr, ok := err.(*googleapi.Error); ok && apiErr.Code == http.StatusNotFound {
+			return false, errors.NewNotFoundError("Host was deleted concurrently", err)
+		}
+		return false, fmt.Errorf("failed to check host existence: %w", err)
+	}
+	return g.basicChecker.IsHostReady()
+}
+
+func (m *GCEInstanceManager) waitForOrchestrator(zone string, host *apiv1.HostInstance, client HostClient) error {
+	gceChecker := &gceHostReadinessChecker{
+		basicChecker: &BasicHostReadinessChecker{Client: client},
+		manager:      m,
+		zone:         zone,
+		hostName:     host.Name,
+	}
+
+	return WaitForHostReady(gceChecker, 5*time.Minute, 5*time.Second)
+}
+
+func (m *GCEInstanceManager) getHostAddrWithIns(ins *compute.Instance) (string, error) {
+	ilen := len(ins.NetworkInterfaces)
+	if ilen == 0 {
+		log.Printf("host instance %s in zone %s is missing a network interface", ins.Name, ins.Zone)
+		return "", errors.NewInternalError("host instance missing a network interface", nil)
+	}
+	if ilen > 1 {
+		log.Printf("host instance %s in zone %s has %d network interfaces", ins.Name, ins.Zone, ilen)
+	}
+	return ins.NetworkInterfaces[0].NetworkIP, nil
+}
+
+func (m *GCEInstanceManager) getHostURLWithIns(ins *compute.Instance) (*url.URL, error) {
+	addr, err := m.getHostAddrWithIns(ins)
+	if err != nil {
+		return nil, err
+	}
+	return url.Parse(fmt.Sprintf("%s://%s:%d", m.Config.HostOrchestratorProtocol, addr, m.Config.GCP.HostOrchestratorPort))
+}
+
+func (m *GCEInstanceManager) getHostClientWithIns(ins *compute.Instance) (HostClient, error) {
+	url, err := m.getHostURLWithIns(ins)
 	if err != nil {
 		return nil, err
 	}
 	return NewNetHostClient(url, m.Config.AllowSelfSignedHostSSLCertificate), nil
+}
+
+func (m *GCEInstanceManager) GetHostClient(zone string, host string) (HostClient, error) {
+	ins, err := m.getHostInstance(zone, host)
+	if err != nil {
+		return nil, err
+	}
+	return m.getHostClientWithIns(ins)
 }
 
 func (m *GCEInstanceManager) getHostInstance(zone string, host string) (*compute.Instance, error) {
@@ -326,6 +407,7 @@ var (
 type opResultGetter struct {
 	Service *compute.Service
 	Op      *compute.Operation
+	Config  *Config
 }
 
 func (g *opResultGetter) Get() (any, error) {
--- a/pkg/app/instances/gce_test.go
+++ b/pkg/app/instances/gce_test.go
@@ -24,6 +24,7 @@ import (
 	"net/http/httptest"
 	"net/url"
 	"reflect"
+	"strconv"
 	"strings"
 	"testing"
 
@@ -572,9 +573,14 @@ func TestWaitCreateInstanceOperationSucceeds(t *testing.T) {
 		Name:           "foo",
 		MachineType:    "mt",
 		MinCpuPlatform: "mcp",
+		NetworkInterfaces: []*compute.NetworkInterface{
+			{NetworkIP: "127.0.0.1"},
+		},
 	}
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch path := r.URL.Path; path {
+		case "/":
+			replyJSON(w, map[string]string{})
 		case "/projects/google.com:test-project/zones/us-central1-a/operations/operation-1/wait":
 			replyJSON(w, operation)
 		case "/projects/google.com:test-project/zones/us-central1-a/instances/foo":
@@ -584,7 +590,18 @@ func TestWaitCreateInstanceOperationSucceeds(t *testing.T) {
 		}
 	}))
 	defer ts.Close()
-	im := NewGCEInstanceManager(testConfig, buildTestService(t, ts), testNameGenerator)
+
+	tsURL, _ := url.Parse(ts.URL)
+	port, _ := strconv.Atoi(tsURL.Port())
+
+	cfg := testConfig
+	cfg.HostOrchestratorProtocol = "http"
+
+	gcpCfg := *cfg.GCP
+	gcpCfg.HostOrchestratorPort = port
+	cfg.GCP = &gcpCfg
+
+	im := NewGCEInstanceManager(cfg, buildTestService(t, ts), testNameGenerator)
 
 	res, _ := im.WaitOperation(zone, &TestUser{}, opName)
 
--- a/pkg/app/instances/helper.go
+++ b/pkg/app/instances/helper.go
@@ -1,0 +1,50 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package instances
+
+import (
+	"net/http"
+	"time"
+
+	"github.com/google/cloud-android-orchestration/pkg/app/errors"
+)
+
+func WaitForHostReady(checker HostReadinessChecker, maxWait time.Duration, retryDelay time.Duration) error {
+	deadline := time.Now().Add(maxWait)
+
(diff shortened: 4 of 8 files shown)


pkg/app/instances/gce.go
```diff
@@ -362,7 +369,19 @@ func (g *opResultGetter) buildCreateInstanceResult() (*apiv1.HostInstance, error
 	if err != nil {
 		return nil, toAppError(err)
 	}
-	return BuildHostInstance(ins)
+
+	host, err := BuildHostInstance(ins)
+	if err != nil {
+		return nil, err
+	}
+	client, err := g.InstanceManager.GetHostClient(g.Zone, host.Name)
```
0405ysj: From `GceIM`, `GetHostClient` internally invokes `getHostInstance`, which seems to be calling GCE API to retrieve the host. However, this method already has `ins` which is typed as `*compute.Instance`.

I think it's better not to get `*compute.Instance` twice, also not to bring new fields(`InstanceManager`, `Zone`) into `opResultGetter`.

k311093: Changed logic to pass `*compute.Instance` and reuse it if specified.

pkg/app/instances/gce.go
```diff
@@ -78,8 +78,16 @@ func (m *GCEInstanceManager) ListZones() (*apiv1.ListZonesResponse, error) {
...
 }
 
-func (m *GCEInstanceManager) GetHostAddr(zone string, host string) (string, error) {
-	instance, err := m.getHostInstance(zone, host)
+func (m *GCEInstanceManager) getHostAddrInner(zone string, host string, ins *compute.Instance) (string, error) {
+	var instance *compute.Instance
+	var err error
+
+	if ins == nil {
+		instance, err = m.getHostInstance(zone, host)
```
0405ysj: I think it's better to expose `getHostInstance` towards testing methods. Then we don't need getHostAddr, getHostAddrInner, getHostURLInner, GetHostURL, getHostURLinner. Instead we can leave them as util functions. HDYT?

k311093: We could change to passing `m.getHostInstance(zone, host)` in `GetHostClient`, and might remove addition functions.

Thanks for the suggestion!

ser-io (changes requested):
Adding an inside logic to wait for when the host is ready to interact with shouldn't be part of the CreateHost request logic on the backend side.

CreateHost returns success when the host is created and host resources are allocated, not when the host is ready to interact with. To do this properly, you'd need to add and implement a new http endpoint that tests whether the host is ready, still clients would need to make these calls after CreateHost returned in order to determine whether the host is ready to interact with. However, it's up to every client implementation how they want to handle this, they can always keep doing what the current client is doing.

pkg/app/instances/gce.go
```diff
@@ -78,8 +78,16 @@ func (m *GCEInstanceManager) ListZones() (*apiv1.ListZonesResponse, error) {
 	}, nil
 }
 
-func (m *GCEInstanceManager) GetHostAddr(zone string, host string) (string, error) {
-	instance, err := m.getHostInstance(zone, host)
+func (m *GCEInstanceManager) getHostAddrInner(zone string, host string, ins *compute.Instance) (string, error) {
```
ser-io: Accepting an argument that can be `nil` and `not nil`  to later do the following is a bad pattern.

```
if ins == nil {
  instance, err = m.getHostInstance(zone, host)
} else {
  instance, err = ins, nil
}
```

These should be two different functions

```
func (m *GCEInstanceManager) GetHostAddr(zone string, host string)
func (m *GCEInstanceManager) GetHostAddrWithIns(ins *compute.Instance)
```

Curious, was this specific change generated by AI?

ser-io (changes requested):

jemoreira:
> Adding an inside logic to wait for when the host is ready to interact with shouldn't be part of the CreateHost request logic on the backend side.
> 
> CreateHost returns success when the host is created and host resources are allocated, not when the host is ready to interact with.

What use is a host that doesn't have a host orchestrator? If the answer is "none", then it would make sense to only mark the operation to create a host as completed once the host orchestrator is ready. It simplifies the api and puts the need for waiting in a single place (the cloud orchestrator) rather than in every client. The CO is already waiting for the host to be created, it might as well wait a bit longer.

jemoreira:
> Context: b/501288123

Please add more details to the description and the title of the PR.

What is "host orchestrator settlement"?

Which backend? The HO or the CO or something else?

Why is this change useful/needed? What are the pros and cons? Even if you write a description in the bug that's not publicly available like this PR is.

pkg/app/instances/gce.go
k311093: No, it was not generated by AI.

I was meant to sharing common code / logic as much as possible, and want to have same log before.

I'll separate functions for zone, host / compute.Instance.

k311093:
> > Context: b/501288123
> 
> Please add more details to the description and the title of the PR.
> 
> What is "host orchestrator settlement"?
> 
> Which backend? The HO or the CO or something else?
> 
> Why is this change useful/needed? What are the pros and cons? Even if you write a description in the bug that's not publicly available like this PR is.

Updated commit message with descriptions.

0405ysj:
> What use is a host that doesn't have a host orchestrator? If the answer is "none"

+1. I expect users are understanding the meaning of  **host** as *a host environment to launch Cuttlefish instances*, and the route of orchestrating CF instances is Host Orchestrator. I think health of Host Orchestrator is more representative than health of computing unit(e.g. GCE or docker instance).

From API view, I expect users create a host and verify with 3 steps; `POST /v1/zones/{zone}/hosts`, `POST /v1/zones/{zone}/operations/{operation}/:wait`, `GET /v1/zones/{zone}/hosts/{host}/`. Invoking those endpoints should be responsible by clients though. But having another round of wait by setting retry count for requesting `GET /v1/zones/{zone}/hosts/{host}/` would be unexpected behavior in users' perspective, I think.

pkg/app/instances/gce.go
k311093: Changed ~~~Inner methods into ~~WithIns, and found that compute.Instance has the fields that can be used as zone, host. So changed methods to pass just *compute.Instance and use its fields.

pkg/app/instances/gce.go
```diff
@@ -324,8 +349,10 @@ var (
 )
 
 type opResultGetter struct {
-	Service *compute.Service
-	Op      *compute.Operation
+	Service         *compute.Service
+	Op              *compute.Operation
+	InstanceManager *GCEInstanceManager
+	Zone            string
```
0405ysj: The field `Zone` doesn't seem to be used?

pkg/app/instances/gce.go
```diff
@@ -253,18 +265,31 @@ func (m *GCEInstanceManager) WaitOperation(zone string, user accounts.User, name
...
+		Op:              op,
+		InstanceManager: m,
+		Zone:            zone,
+	}
 	return getter.Get()
 }
 
-func (m *GCEInstanceManager) GetHostClient(zone string, host string) (HostClient, error) {
-	url, err := m.GetHostURL(zone, host)
+func (m *GCEInstanceManager) GetHostClientWithIns(ins *compute.Instance) (HostClient, error) {
```
0405ysj: Can we pass `m.Config.AllowSelfSignedHostSSLCertificate` as a parameter of `GetHostClientWithIns`? `GetHostAddrWithIns` and `GetHostURLWithIns` don't seem to be methods of `*GCEInstanceManager`. When we pass it as a parameter, `GetHostClientWithIns` doesn't need to be a method of it either. Then we don't need to define field `InstanceManager` on `opResultGetter`.

k311093: Changed ...Ins methods to internal functions that receives instance, config. and changed to use that functions. Looks clearer than before. Thanks!

pkg/app/instances/gce.go
k311093: Removed unused field while changing code.

pkg/app/instances/gce.go
```diff
@@ -324,8 +348,9 @@ var (
 )
 
 type opResultGetter struct {
-	Service *compute.Service
-	Op      *compute.Operation
+	Service         *compute.Service
+	Op              *compute.Operation
+	InstanceManager *GCEInstanceManager
```
0405ysj: Config instead?

0405ysj (approved):

pkg/app/instances/gce.go
k311093: Changed to config.

pkg/client/client.go
```diff
@@ -126,19 +126,6 @@ func (c *clientImpl) CreateHost(req *apiv1.CreateHostRequest) (*apiv1.HostInstan
 		return nil, err
 	}
 
-	// There is a short delay between the creation of the host and the availability of the host
```
ser-io, line 129: Don't remove this code in the same PR (commit) the backend code is changed. Add a TODO to remove this code a few weeks later, once you're confident the deployed backends have now the wait logic in-place.

If the client code gets deployed first than the backend code, clients are going to start failing.

ser-io (changes requested):

pkg/app/instances/gce.go
```diff
@@ -362,7 +387,19 @@ func (g *opResultGetter) buildCreateInstanceResult() (*apiv1.HostInstance, error
...
+
+	host, err := BuildHostInstance(ins)
+	if err != nil {
+		return nil, err
+	}
+	client, err := getHostClientWithIns(ins, g.Config)
+	if err != nil {
+		return nil, err
+	}
+	if err := client.WaitForHostReady(); err != nil {
```
ser-io: Do not wait on GET requests. In CO and HO we have POST /:wait endpoints for long running/blocking operations. In this case you'd need to add the wait logic to the POST /:wait logic relevant for creating the host.

ser-io (changes requested):

ser-io:
> > Adding an inside logic to wait for when the host is ready to interact with shouldn't be part of the CreateHost request logic on the backend side.
> > CreateHost returns success when the host is created and host resources are allocated, not when the host is ready to interact with.
> 
> What use is a host that doesn't have a host orchestrator? If the answer is "none", then it would make sense to only mark the operation to create a host as completed once the host orchestrator is ready. It simplifies the api and puts the need for waiting in a single place (the cloud orchestrator) rather than in every client. The CO is already waiting for the host to be created, it might as well wait a bit longer.

Right now the backend doesn't care whether the HO is running or not, it just create/list/delete hosts. Then, waiting to check whether you can interact with the HO is passed down to the client implementation, which I think makes things more simple conceptually.

Moving this responsibility to the backend makes the backend now aware of a new WAITING_FOR_HO status.

How do you plan to list hosts: are you listing all hosts or only listing hosts with the HO ready? If a host was created successfully  but the OS failed to boot or the HO failed to launch, as an infra solution, I still want to see that host listed for triaging purposes.

k311093:
> > > Adding an inside logic to wait for when the host is ready to interact with shouldn't be part of the CreateHost request logic on the backend side.
> > > CreateHost returns success when the host is created and host resources are allocated, not when the host is ready to interact with.
> > 
> > 
> > What use is a host that doesn't have a host orchestrator? If the answer is "none", then it would make sense to only mark the operation to create a host as completed once the host orchestrator is ready. It simplifies the api and puts the need for waiting in a single place (the cloud orchestrator) rather than in every client. The CO is already waiting for the host to be created, it might as well wait a bit longer.
> 
> Right now the backend doesn't care whether the HO is running or not, it just create/list/delete hosts. Then, waiting to check whether you can interact with the HO is passed down to the client implementation, which I think makes things more simple conceptually.
> 
> Moving this responsibility to the backend makes the backend now aware of a new WAITING_FOR_HO status.
> 
> How do you plan to list hosts: are you listing all hosts or only listing hosts with the HO ready? If a host was created successfully but the OS failed to boot or the HO failed to launch, as an infra solution, I still want to see that host listed for triaging purposes.

To handle this and comment about gce.go, I tried to add task manager, but it make code more complex than expected and it this is not beneficial than introducing complexity. It might be better to handle that in the client client reduces the complexity. How about just add instruction for comment in `CreateHostRequest` `api/v1/instancemanager.go` ?

ser-io:
> > > > Adding an inside logic to wait for when the host is ready to interact with shouldn't be part of the CreateHost request logic on the backend side.
> > > > CreateHost returns success when the host is created and host resources are allocated, not when the host is ready to interact with.
> > > 
> > > 
> > > What use is a host that doesn't have a host orchestrator? If the answer is "none", then it would make sense to only mark the operation to create a host as completed once the host orchestrator is ready. It simplifies the api and puts the need for waiting in a single place (the cloud orchestrator) rather than in every client. The CO is already waiting for the host to be created, it might as well wait a bit longer.
> > 
> > 
> > Right now the backend doesn't care whether the HO is running or not, it just create/list/delete hosts. Then, waiting to check whether you can interact with the HO is passed down to the client implementation, which I think makes things more simple conceptually.
> > Moving this responsibility to the backend makes the backend now aware of a new WAITING_FOR_HO status.
> > How do you plan to list hosts: are you listing all hosts or only listing hosts with the HO ready? If a host was created successfully but the OS failed to boot or the HO failed to launch, as an infra solution, I still want to see that host listed for triaging purposes.
> 
> To handle this and comment about gce.go, I tried to add task manager, but it make code more complex than expected and it this is not beneficial than introducing complexity. It might be better to handle that in the client client reduces the complexity. How about just add instruction for comment in `CreateHostRequest` `api/v1/instancemanager.go` ?

Comment looks good to me.

jemoreira:
> Right now the backend doesn't care whether the HO is running or not, it just create/list/delete hosts. Then, waiting to check whether you can interact with the HO is passed down to the client implementation, which I think makes things more simple conceptually.

It makes the CO implementation simpler, not its API surface. That's the opposite of what we should aim for.

> Moving this responsibility to the backend makes the backend now aware of a new WAITING_FOR_HO status.

The backend is aware of it, but the client is not. That's just a tradeoff, and a worth it one given that there can be multiple clients and single backend.

> How do you plan to list hosts: are you listing all hosts or only listing hosts with the HO ready? If a host was created successfully but the OS failed to boot or the HO failed to launch, as an infra solution, I still want to see that host listed for triaging purposes.

How do you list hosts that are currently being created? A host without a host orchestrator ready is simply not fully "created" yet. You could add a timeout to that wait and fail the operation after that timeout too.

pkg/app/instances/gce.go
jemoreira: +1

Waiting for the host to be ready (HO is up and running) should be done in the wait-for-create-operation-to-complete endpoint. Also, please take into account the following caveats:
* The /:wait endpoint isn't just for waiting for host creation to complete, it's used to wait for all types of operations. You're going to need to differentiate between the different operation types to only add the extra wait to the correct operation. You'll also need a way to figure out the host name from the operation.
* Waiting for the HO in the CO should have a timeout, after which the create operation should return failure. The timeout should depend on the actual InstanceManager implementation, as GCE hosts boot times are different than docker containers'.
* The returned failure be clear enough to differentiate between a failure to create the host and the HO not becoming reachable before the timeout.
* The wait for operation must detect and handle correctly (return failure) if a host was successfully created but deleted by a concurrent call before the create operation succeeded (it's ok to wait until the timeout the first time, but any subsequent wait should return immediately; this means in practices that you should check the host exists before you try to contact the HO).

While the wait-for-create-operation is being called, the list operation may return the existing host, even when the HO isn't ready in it. That's the way it works today, it's fine to keep that behavior for now.

jemoreira (changes requested):

ser-io:
> > > > > Adding an inside logic to wait for when the host is ready to interact with shouldn't be part of the CreateHost request logic on the backend side.
> > > > > CreateHost returns success when the host is created and host resources are allocated, not when the host is ready to interact with.
> > > > 
> > > > 
> > > > What use is a host that doesn't have a host orchestrator? If the answer is "none", then it would make sense to only mark the operation to create a host as completed once the host orchestrator is ready. It simplifies the api and puts the need for waiting in a single place (the cloud orchestrator) rather than in every client. The CO is already waiting for the host to be created, it might as well wait a bit longer.
> > > 
> > > 
> > > Right now the backend doesn't care whether the HO is running or not, it just create/list/delete hosts. Then, waiting to check whether you can interact with the HO is passed down to the client implementation, which I think makes things more simple conceptually.
> > > Moving this responsibility to the backend makes the backend now aware of a new WAITING_FOR_HO status.
> > > How do you plan to list hosts: are you listing all hosts or only listing hosts with the HO ready? If a host was created successfully but the OS failed to boot or the HO failed to launch, as an infra solution, I still want to see that host listed for triaging purposes.
> > 
> > 
> > To handle this and comment about gce.go, I tried to add task manager, but it make code more complex than expected and it this is not beneficial than introducing complexity. It might be better to handle that in the client client reduces the complexity. How about just add instruction for comment in `CreateHostRequest` `api/v1/instancemanager.go` ?
> 
> Comment looks good to me.

Let's do https://github.com/google/cloud-android-orchestration/pull/521#discussion_r3120792871

k311093:
> Let's do https://github.com/google/cloud-android-orchestration/pull/521#discussion_r3120792871

Got it. Thanks!

k311093:
Added endpoint `:wait-host-availability` endpoint and changed `cvdr` to use it. When `:wait-host-availability` fails, fallback to the logic used before.

pkg/app/app.go
```diff
@@ -106,6 +106,7 @@ func (c *App) Handler() http.Handler {
 	// `Get`/`Create`/`Update`, the response should be the relevant resource.
 	router.Handle("/v1/zones/{zone}/operations/{operation}/:wait", c.Authenticate(c.waitOperation)).Methods("POST")
 	router.Handle("/v1/zones/{zone}/hosts/{host}", c.Authenticate(c.deleteHost)).Methods("DELETE")
+	router.Handle("/v1/zones/{zone}/hosts/{host}/:wait-host-availability", c.Authenticate(c.waitCreateHost)).Methods("POST")
```
ser-io: The CO API is designed to have only one `:wait` endpoint `/v1/zones/{zone}/operations/{operation}/:wait`, there are no specific wait endpoints like:  /:wait-for-create or /:wait-for-delete for example.

For the purpose of this PR specifically wondering whether can you can extend the wait logic when the operation was a **create host operation** and wait a bit longer till the HO is reachable rather than adding a new endpoint.

ser-io (changes requested):

pkg/app/app.go
k311093: In that case, creating another endpoint for creating operation for waiting host orchestrator's settlement and wait it using previous :wait would be a proper way?

ser-io: The purpose of this PR was to simplify the client logic. Creating a new operation for waiting on HO readiness doesn't simplify the client implementation, it makes things even more complicated than what we have right now.

The way to make client logic simpler is for the client to make two calls only: 

```
POST /host
POST /operation/foo/:wait
```

When the operation was a "create host operation". `POST /operation/foo/:wait` should be resolved when the host was created **and the HO is ready**. Right now we resolve "create host operation" the moment the host was created without waiting for the HO to be ready, we need to extend this waiting logic for create host operations and not add a new endpoint.

k311093: Changed to wait for host ready at `:wait` like other operations. Moved host wait logic to `WaitOperation` function, not `Get()` method like before.

Checked with cloud orchestrator at GCP, docker and confirmed that it successfully created host and subsequent tasks (for GCP side, didn't checked CVD creation completely, but checked that uploading, extracting artifacts worked well).

ser-io: If the wait logic was moved behind `POST /v1/zones/{zone}/operations/{operation}/:wait` can we remove `POST /v1/zones/{zone}/hosts/{host}/:wait-host-availability`?

ser-io (changes requested):

pkg/app/app.go
k311093: I missed to remove `:wait-host-availability`, removed that endpoint and also changed related codes.

Meanwhile, I found that waiting logic is not common for docker and GCP, separated its own logic to internal codes.

Also updated tests for each instance manager.

pkg/app/instances/docker.go
```diff
@@ -188,6 +189,13 @@ func (m *DockerInstanceManager) waitCreateHostOperation(host string) (*apiv1.Hos
 				return nil, fmt.Errorf("failed to inspect docker container: %w", err)
 			}
 			if res.State.Running {
+				client, err := m.GetHostClient("local", host)
+				if err != nil {
+					return nil, err
+				}
+				if err := waitForHostReady(client); err != nil {
```
ser-io: Add a comment mentioning that there's a delay between the host is created and the HO is ready, and we want to return once the HO is ready.

ser-io: Pass "maxWait" and "retryDelay" so we don't have to look into the body of the `waitForHostReady` function to understand how long the wait could be.

pkg/app/instances/docker.go
```diff
@@ -500,3 +508,18 @@ func (m *DockerInstanceManager) getRWMutex(user accounts.User) *sync.RWMutex {
 	mu, _ := m.mutexes.LoadOrStore(user.Username(), &sync.RWMutex{})
 	return mu.(*sync.RWMutex)
 }
+
+func waitForHostReady(client HostClient) error {
```
ser-io: Make this a public reusable helper  to be used among other instance managers. `WaitForHostReady`. Have them receive  a smaller interface with a PingHO() method or similar rather than passing a full Client implementation.

pkg/app/instances/gce.go
ser-io, line 1: Let's split this PR into two different commits. One for docker.go and one for gce.go.

pkg/app/instances/gce.go
```diff
@@ -41,6 +43,7 @@ type GCPIMConfig struct {
 	UseExternalIP        bool
 	// If true, instances created should be compatible with `acloud CLI`.
 	AcloudCompatible bool
+	HostReadyTimeout time.Duration
```
ser-io: Let's remove this config in this PR and use good hardcoded defaults as we have for docker. We can add the config if needed in the future.

pkg/app/instances/gce.go
```diff
@@ -78,28 +81,40 @@ func (m *GCEInstanceManager) ListZones() (*apiv1.ListZonesResponse, error) {
 	}, nil
 }
 
-func (m *GCEInstanceManager) GetHostAddr(zone string, host string) (string, error) {
-	instance, err := m.getHostInstance(zone, host)
-	if err != nil {
-		return "", err
-	}
-	ilen := len(instance.NetworkInterfaces)
+func getHostAddrWithIns(ins *compute.Instance) (string, error) {
```
ser-io: Move these new set of helper functions as a continuations block in the end of the file `getHostAddrWithIns`, `getHostURLWithIns` to the end. So methods like `func (m *GCEInstanceManager) GetHostAddr` and `func (m *GCEInstanceManager) GetHostURL` continued to be grouped in the same block.

pkg/app/instances/gce.go
```diff
@@ -253,16 +268,81 @@ func (m *GCEInstanceManager) WaitOperation(zone string, user accounts.User, name
...
+
+func (m *GCEInstanceManager) waitForOrchestrator(zone string, host *apiv1.HostInstance, client HostClient) error {
+	timeout := m.Config.GCP.HostReadyTimeout
+	if timeout == 0 {
+		timeout = 5 * time.Minute
+	}
+	retryDelay := 5 * time.Second
+	deadline := time.Now().Add(timeout)
+
+	for time.Now().Before(deadline) {
```
ser-io: Reuse the wait logic implemented in docker.go.

pkg/app/instances/gce_test.go
```diff
@@ -594,6 +612,104 @@ func TestWaitCreateInstanceOperationSucceeds(t *testing.T) {
...
+	if err == nil {
+		t.Error("expected error")
+	}
+	var appErr *apperr.AppError
+	if errors.As(err, &appErr) {
+		if appErr.StatusCode != http.StatusNotFound {
+			t.Errorf("expected status 404, got %d", appErr.StatusCode)
+		}
+	} else {
+		t.Errorf("expected AppError, got %T", err)
```
ser-io: Remove these unit tests.  Have only unit tests for the new helper function that would be used for Docker and Gce.

ser-io (changes requested):

pkg/app/instances/docker.go
k311093: Added comment about mentioning HO's readiness and need to wait for it.

k311093: Changed function to receive timeout and delay directly.

pkg/app/instances/docker.go
k311093: Made waitForHostReady to common function in helper.go, also added tests about it. Added simple interface `HostChecker` which has `PingHO()`.

pkg/app/instances/gce.go
k311093: Splitted commit into two parts, Docker with common implementations about Waiting logic, and then apply it to CGE logic.

pkg/app/instances/gce.go
k311093: Changed it to use literal time values instead of configuration.

pkg/app/instances/gce.go
k311093: Moved function near its context to make it more easy read.

pkg/app/instances/gce.go
k311093: Changed to use common logic which is at `helper.go`

pkg/app/instances/gce_test.go
k311093: Removed tests because WaitForHostReady tested separated helper test.

pkg/app/instances/instances.go
```diff
@@ -38,7 +38,12 @@ type Manager interface {
 	GetHostClient(zone string, host string) (HostClient, error)
 }
 
+type HostChecker interface {
+	PingHO() (int, error)
+}
+
 type HostClient interface {
+	HostChecker
```
ser-io: `WaitForHostReady` receives a `HostChecker` implementation. We don't need to force every `HostClient` implementation to implement `HostChecker`.

pkg/app/instances/instances.go
```diff
@@ -38,7 +38,12 @@ type Manager interface {
 	GetHostClient(zone string, host string) (HostClient, error)
 }
 
+type HostChecker interface {
```
ser-io: Let's use a more specific name to keep this interface smaller: `HostReadinessChecker`.

pkg/app/instances/instances.go
```diff
@@ -38,7 +38,12 @@ type Manager interface {
...
+	PingHO() (int, error)
```
ser-io: We can have something like: `IsHostReady() (bool, error)` and let every implementation decides. This allow us to remove the `check func() error` argument in `WaitForHostReady` function and have have that logic in the specific IsHostReady implementation. 

// Return `true, nil` if host is ready. 
// Return `false, nil` host is not yet ready, you can retry.
// Return `false, error` readiness cannot be verified, fail and propagate the error.
IsHostReady() (bool, error)

ser-io (changes requested):

pkg/app/instances/instances.go
k311093: Separated Check and Client interface.

pkg/app/instances/instances.go
k311093: Changed name to HostReadinessChecker.

pkg/app/instances/instances.go
k311093: Changed to `IsHostReady() (bool, error)` and use default implementation for Docker instance manager, for GCE instance manager, use special HostChecker which checks host deletion for `IsHostReady`.

pkg/app/instances/helper.go
```diff
@@ -0,0 +1,50 @@
...
+package instances
+
+import (
+	"net/http"
+	"time"
+
+	"github.com/google/cloud-android-orchestration/pkg/app/errors"
+)
+
+func WaitForHostReady(client ReadinessChecker, maxWait time.Duration, retryDelay time.Duration) error {
```
ser-io: nit: `checker`, not `client`.

pkg/app/instances/instances.go
```diff
@@ -38,6 +38,10 @@ type Manager interface {
 	GetHostClient(zone string, host string) (HostClient, error)
 }
 
+type ReadinessChecker interface {
```
ser-io: nit: `HostReadinessChecker`

pkg/app/instances/helper.go
```diff
@@ -0,0 +1,50 @@
...
+		}
+		if ready {
+			return nil
+		}
+		time.Sleep(retryDelay)
+	}
+	return errors.NewServiceUnavailableError("wait for host orchestrator timed out", nil)
+}
+
+type defaultReadinessChecker struct {
```
ser-io: `basicHostReadinessChecker` or `simpleHostReadinessChecker`.

pkg/app/instances/helper_test.go
```diff
@@ -0,0 +1,93 @@
...
+package instances
+
+import (
+	"net/http"
+	"net/http/httputil"
+	"testing"
+	"time"
+)
+
+type mockHostClient struct {
```
ser-io: Why do we need a mock client? We can have a HostReadinessChecker mock for testing `WaitForHostReady`.

pkg/app/instances/helper.go
ser-io: make this basic checker implementation public., `type BasicHostReadinessChecker struct`

pkg/app/instances/gce.go
```diff
@@ -253,16 +247,108 @@ func (m *GCEInstanceManager) WaitOperation(zone string, user accounts.User, name
...
+	if err != nil {
+		return nil, err
+	}
+	if err := m.waitForOrchestrator(zone, hostInstance, client); err != nil {
+		return nil, err
+	}
+	return hostInstance, nil
+}
+
+type gceReadinessChecker struct {
```
ser-io: nit :`gceHostReadinessChecker`

pkg/app/instances/gce.go
```diff
@@ -253,16 +247,108 @@ func (m *GCEInstanceManager) WaitOperation(zone string, user accounts.User, name
...
+	checker  ReadinessChecker
```
ser-io: Pass the BasicHostReadinessChecker directly, which is what we want and not  any implementation of that interface

ser-io (changes requested):

pkg/app/instances/helper.go
k311093: Changed name to `checker`.

pkg/app/instances/helper.go
k311093: Changed to `BasicHostReadinessChecker` which is public.

pkg/app/instances/instances.go
k311093: Changed to `HostReadinessChecker`.

pkg/app/instances/helper_test.go
k311093: Created and changed to use `HostReadinessChecker` for making test simpler.

pkg/app/instances/helper.go
k311093: Changed `defaultReadinessChecker` to public `BasicHostReadinessChecker`, and also changed its member variable `client` to `Client` to make it public accessible.

pkg/app/instances/gce.go
k311093: Changed to `gceHostReadinessChecker`.

pkg/app/instances/gce.go
k311093: Changed to use `BasicHostReadinessChecker` directly for `gceReadinessChecker`.

pkg/app/instances/helper.go
```diff
@@ -0,0 +1,50 @@
...
+		ready, err := checker.IsHostReady()
+		if err != nil {
+			return err
+		}
+		if ready {
+			return nil
+		}
+		time.Sleep(retryDelay)
+	}
+	return errors.NewServiceUnavailableError("wait for host orchestrator timed out", nil)
```
ser-io: Let's  now return `NewInternalError` after time out, otherwise the client will retry again.

pkg/app/instances/gce.go
```diff
@@ -79,27 +81,19 @@ func (m *GCEInstanceManager) ListZones() (*apiv1.ListZonesResponse, error) {
...
 }
 
 func (m *GCEInstanceManager) GetHostURL(zone string, host string) (*url.URL, error) {
-	addr, err := m.GetHostAddr(zone, host)
+	ins, err := m.getHostInstance(zone, host)
 	if err != nil {
 		return nil, err
 	}
-	return url.Parse(fmt.Sprintf("%s://%s:%d", m.Config.HostOrchestratorProtocol, addr, m.Config.GCP.HostOrchestratorPort))
+	return getHostURLWithIns(ins, &m.Config)
```
ser-io: Can you have getHostURLWithIns as a method of `GCEInstanceManager` in order to reuse its `Config` field rather than have it to pass as a parameter? Same for other similar functions in this PR

ser-io (changes requested):

pkg/app/instances/helper.go
k311093: Changed to return error using NewInternalError.

pkg/app/instances/gce.go
k311093: Changed `getHostURLWithIns` like functions to member method if `GCEInstanceManager` to reuse `Config`.

pkg/app/instances/docker.go
```diff
@@ -188,6 +188,18 @@ func (m *DockerInstanceManager) waitCreateHostOperation(host string) (*apiv1.Hos
 				return nil, fmt.Errorf("failed to inspect docker container: %w", err)
 			}
 			if res.State.Running {
+				client, err := m.GetHostClient("local", host)
+				if err != nil {
+					return nil, err
+				}
+				// There is a delay between host creation and its readiness, wait for host readiness and return when it is ready.
+				readinessChecker, ok := client.(HostReadinessChecker)
```
ser-io: let's just have `readinessChecker := &BasicHostReadinessChecker{Client: client}` and remove the casting logic.

pkg/app/instances/helper.go
```diff
@@ -0,0 +1,50 @@
...
+		ready, err := checker.IsHostReady()
+		if err != nil {
+			return err
+		}
+		if ready {
+			return nil
+		}
+		time.Sleep(retryDelay)
+	}
+	return errors.NewInternalError("wait for host orchestrator timed out", nil)
```
ser-io: Let's use a more specific error message: "wait for host orchestrator readiness timed out"

pkg/app/instances/gce.go
```diff
@@ -253,18 +247,105 @@ func (m *GCEInstanceManager) WaitOperation(zone string, user accounts.User, name
...
+	}
+	if hostInst, ok := res.(*apiv1.HostInstance); ok && op.OperationType == "insert" {
+		return m.WaitHostAvailability(zone, user, hostInst.Name)
+	}
+	return res, nil
 }
 
-func (m *GCEInstanceManager) GetHostClient(zone string, host string) (HostClient, error) {
-	url, err := m.GetHostURL(zone, host)
+func (m *GCEInstanceManager) WaitHostAvailability(zone string, user accounts.User, host string) (*apiv1.HostInstance, error) {
```
ser-io: do this new method needs to be public?

ser-io (approved):

pkg/app/instances/docker.go
k311093: Changed to just use `BasicHostReadinessChecker` instead of casting.

pkg/app/instances/helper.go
k311093: Modified error message more specific.

pkg/app/instances/gce.go
k311093: It noes not needs to be, change to `waitHostAvailability` to be private.

k311093:
Hi @jemoreira can you please re-review this PR since request change is not closed yet.

PR #525, opened 2026-05-13
ci: grant contents: read explicitly on presubmit and deploy entry workflows

arpitjain099:
Three workflow files declare no `permissions:` block. `deployment.yaml` (the reusable one called by `postsubmit` and `unstable-channel`) already has an explicit `permissions: contents: read`, but the three entry workflows still rely on the repo default.

- `presubmit.yaml` — build + test + format + vet + staticcheck + Debian/Docker artifact builds + E2E tests. All read-side from a GITHUB_TOKEN perspective.
- `postsubmit.yaml` and `unstable-channel.yaml` — both just delegate to `deployment.yaml` with secrets. Adding `contents: read` here intersects with the reusable`s declared `contents: read` = same scope, no regression. The `ARTIFACT_REGISTRY_UPLOADER` secret is what authenticates the actual GCP push.

YAML parses cleanly.

### DIFF
--- a/.github/workflows/postsubmit.yaml
+++ b/.github/workflows/postsubmit.yaml
@@ -5,6 +5,9 @@ on:
     branches:
       - main
 
+permissions:
+  contents: read
+
 jobs:
   deploy:
     if: github.repository_owner == 'google'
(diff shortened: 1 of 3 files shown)


0405ysj:
Thanks for the contribution, and clear description in PR! Approved the PR.

0405ysj (approved):

PR #532, opened 2026-06-03
Use adb serial to assert adb connection in e2e tests.

ser-io:


e2etests/cvdr/cvdr_create_with_branch_test.sh
```diff
@@ -13,11 +13,16 @@ cleanup() {
...
-    --host=${HOSTNAME} \
-    --branch=aosp-android-latest-release \
-    --build_target=aosp_cf_x86_64_only_phone-userdebug
+out=$(cvdr create \
+  --host=${HOSTNAME} \
+  --branch=aosp-android-latest-release \
+  --build_target=aosp_cf_x86_64_only_phone-userdebug)
+
+# Verify adb connection
+adb_serial=$(echo ${out} | grep 'ADB:' | grep -oP '127.0.0.1:\d+')
```
jemoreira: nit: Not long ago the serial was `0.0.0.0:{port}`, it could change to be `localhost:{port}` or something like `vsock:{cid}:{port}`. Consider making that last grep more general, something like `grep -oE '[^ ]+$'`

e2etests/cvdr/cvdr_create_with_local_srcs_test.sh
```diff
@@ -24,11 +24,15 @@ cleanup() {
...
     --host=${HOSTNAME} \
     --local_cvd_host_pkg_src=${CVD_HOST_PKG} \
-    --local_images_zip_src=${IMAGE_ZIP}
+    --local_images_zip_src=${IMAGE_ZIP})
 
-# Check ADB connection
-# TODO(b/448209030): Retrieve serial of the device from the output of cvdr list.
-adb shell uptime
+# Verify adb connection
+adb_serial=$(echo ${out} | grep 'ADB:' | grep -oP '127.0.0.1:\d+')
```
jemoreira: is there a way to encapsulate this so it doesn't have to be repeated in every test?

jemoreira (approved):

e2etests/cvdr/cvdr_create_with_branch_test.sh
ser-io: Good point! Done.

e2etests/cvdr/cvdr_create_with_local_srcs_test.sh
ser-io: Done. Moved the logic to common_utils.sh.

PR #535, opened 2026-07-14
Avoid overlayfs on /var/tmp/cvd and /tmp not to be blocked by disk I/O

0405ysj:
Context: b/533199674


0405ysj:
Closing this PR, as this didn't seem to be the root cause of the bug in the context.

